PR #89, opened 2020-01-27
Broadcast transformations

bbalser:


apps/service_broadcast/lib/broadcast/stream/broadway.ex
```diff
@@ -51,11 +60,25 @@ defmodule Broadcast.Stream.Broadway do
     messages
   end
 
-  defp setup_config(load) do
+  defp create_transformer(dataset_id) do
+    case Broadcast.Transformations.get(dataset_id) do
+      {:ok, nil} ->
+        fn x -> Ok.ok(x) end |> Ok.ok()
```
jeffgrunewald: what's the double `Ok.ok/1` buying you here?

bbalser: create transformer need to return an {:ok, function} and the function needs to return a ok tuple

jeffgrunewald: 👍

PR #90, opened 2020-01-27
Removed normalization from persist and added transformation

bbalser:


apps/service_persist/lib/persist/loader.ex
```diff
@@ -24,11 +26,18 @@ defmodule Persist.Loader do
...
+    with {:ok, transform} when not is_nil(transform) <- Transformations.get(load.dataset_id),
+         {:ok, dictionary} <- transform_dictionary(transform),
+         {:ok, writer_pid} <- start_writer(load, dictionary),
+         writer_function <- fn msgs ->
+           writer().write(writer_pid, msgs, dictionary: dictionary)
+         end,
+         {:ok, broadway_pid} <- start_broadway(load, transform, writer_function) do
       {:ok, %{writer_pid: writer_pid, broadway_pid: broadway_pid}}
     else
+      {:ok, nil} ->
```
jeffgrunewald: since you said there always needs to be a transform, should it assume a default `:noop` transform if it doesn't find one?

apps/service_persist/test/persist/load/broadway_test.exs
```diff
@@ -67,60 +82,21 @@ defmodule Persist.Load.BroadwayTest do
...
-        dataset_id: "ds1",
-        name: "fake-name",
-        source: "topic-c",
-        destination: "table-a",
-        schema: [
-          %Dictionary.Type.String{name: "name"},
-          %Dictionary.Type.Integer{name: "age"}
-        ]
-      )
+    # load =
```
jeffgrunewald: do these need to be uncommented or deleted?

apps/service_persist/lib/persist/loader.ex
bbalser: I think stopping the process is best, because if now transform is found then we also don't know the dictionary and create the table in presto.

PR #92, opened 2020-01-28
fixing readme for properties

jeffgrunewald:
I don't know how you'd do this accidentally, so I assume it was done on purpose to irritate my OCD.

### DIFF
--- a/apps/properties/README.md
+++ b/apps/properties/README.md
@@ -1,25 +1,38 @@
 # Properties
 
-Asynchronously check test outcomes that are expected to
-validate but may not immediately.
+Properties provides a series of macros for accessing
+application configuration (properties) within your modules.
 
-AssertAsync implements a macro that injects the necessary
-retry logic for validating test assertions with a configurable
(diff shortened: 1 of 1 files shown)


apps/properties/README.md
```diff
@@ -1,25 +1,38 @@
...
 
 ## Usage
 
 ```elixir
   defmodule Example do
-    use ExUnit.Case
-    import AssertAsync
+    use Properties, otp_app: :example_service
+
+    @app_name get_config_value(:app_name, required: true)
```
bbalser: Should we be clear about not using this format unless absolutely necessary?

jeffgrunewald: which format specifically? using get_config_value and binding it to a module attribute?

bbalser: yeah, the goal is to make it painless to access the config at runtime and get_config_value should only be used when the config has to be done a compile time.

jeffgrunewald: sure thing.

PR #105, opened 2020-02-05
Receive udp

jeffgrunewald:
Ready to start getting feedback on the udp socket

Still need to plumb out the interactions of the actual service and probably extract the `test/support` stuff for accept_udp into the testing library to allow pulling the source socket into the platform runner/e2e tests as well.


apps/service_receive/lib/receive/acception/store.ex
```diff
@@ -0,0 +1,33 @@
+defmodule Receive.Acception.Store do
+  @instance Receive.Application.instance()
+  @collection "acceptions"
```
jdenen: This... hurts to read. It just, is not a word that should ever exist.

jdenen: `accepting`? `acceptables`?

.tool-versions
```diff
@@ -1,3 +1,4 @@
 erlang 22.2.1
+erlang 22.2.4
```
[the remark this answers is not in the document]
jeffgrunewald: nope

apps/service_receive/lib/receive/acception/store.ex
jeffgrunewald: changed to `accepts`

jeffgrunewald:
Well played, @jdenen

PR #106, opened 2020-02-06
Dockerize

jdenen:


.dockerignore
```diff
@@ -0,0 +1,2 @@
+_build/
+deps/
```
jeffgrunewald: do you need to explicitly include any of the `.git` directories or does dockerignore file assume that one by default?
might want to include some of the files/extensions for helm too, unless you plan to put the charts in a different repo.

Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM bitwalker/alpine-elixir-phoenix:1.9.4 as build
+COPY . /opt/app
+WORKDIR /opt/app
+RUN mix do local.hex --force, local.rebar --force, deps.get --only prod, deps.compile
+ENV MIX_ENV=prod
+RUN mix release orchestrate \
```
jeffgrunewald: does the same `mix do *` trick not work for the release command?

Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+  && mix release gather \
+  && mix release broadcast \
+  && mix release persist \
+  && mix release acquire
+
+FROM bitwalker/alpine-erlang:22.2.3
+WORKDIR /opt/app
+COPY --from=build /opt/app/_build/prod/rel/orchestrate/ .
```
jeffgrunewald: not sure if there's any better performance or readability to it but would it help to do a single `COPY` of the `rel` directory and then do a `mkdir/mv` within a subsequent `RUN` command?

Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+WORKDIR /opt/app
+RUN mix do local.hex --force, local.rebar --force, deps.get --only prod, deps.compile
+ENV MIX_ENV=prod
+RUN mix release orchestrate \
+  && mix release gather \
+  && mix release broadcast \
+  && mix release persist \
+  && mix release acquire
+
+FROM bitwalker/alpine-erlang:22.2.3
```
jeffgrunewald: I assume this is smaller but do we miss out on any debugging tooling by not running these in the `alpine-elixir` image?

config/releases.exs
```diff
@@ -22,6 +22,12 @@ config :logger, :console,
 # Use Jason for JSON parsing in Phoenix
 config :phoenix, :json_library, Jason
 
+secret_key_base =
+  System.get_env(
+    "SECRET_KEY_BASE",
+    "d2cgmPzW+bqVjs99FUeKJ0kOm0w8EZBvLS7UBM8EHi6uBKgW2oBAa9pR2KSu8Z2W"
+  )
+
 presto_db = [url: "http://localhost:8080", catalog: "hive", schema: "default"]
```
jeffgrunewald: do we set ourselves up for success by making this presto_db url already a `System.get_env("PRESTO_URL" || "localhost")`?

config/releases.exs
```diff
@@ -216,4 +224,13 @@ config :service_orchestrate, Orchestrate.Application,
   ]
 
 # SERVICE ACQUIRE
-config :service_acquire, Acquire.Db.Presto, presto: Keyword.put(presto_db, :user, "acquire")
+config :service_acquire, AcquireWeb.Endpoint,
+  http: [:inet6, port: String.to_integer(System.get_env("PORT") || "4000")],
+  url: [host: "localhost"],
```
jeffgrunewald: Same `System.get_env` callout here as far as the default for the hostname?

config/releases.exs
jdenen: Yeah definitely should be getting this value from `System`. Nice catch.

config/releases.exs
jdenen: This is just used when generating URLs from the backend. Looks like I should remove the config entirely to remove potential confusion.

Dockerfile
jdenen: I was having trouble with it earlier, but I think that was because of a `.dockerignore` issue. I'll take another look.

Dockerfile
jdenen: I'll play with this. It might be better, yeah.

config/releases.exs
jeffgrunewald: 👍

Dockerfile
jdenen: Nope, `mix do` and `mix release` do not play well together.

Dockerfile
jdenen: I don't think so. I think we'll have more available to us than the `smartcitiesdata` images have.

Dockerfile
jeffgrunewald: fair enough

Dockerfile
jeffgrunewald: 👍

PR #107, opened 2020-02-07
Change name to subset id and ensure used a . key to view states

bbalser:


apps/platform_runner/test/performance_test.exs
```diff
@@ -52,12 +53,10 @@ defmodule Platform.Runner.PerformanceTest do
...
         dataset_id: "perf-csv-ds",
-        dictionary: [
-          Dictionary.Type.String.new!(name: "letter"),
-          Dictionary.Type.String.new!(name: "number")
-        ],
+        subset_id: "default",
+        dictionary: dictionary,
         steps: [
-          Transformer.MoveField.new!(from: "letter", to: "single_letter")
+          # Transformer.MoveField.new!(from: "letter", to: "single_letter")
```
jdenen: Did you want to keep this or uncomment it?

bbalser: it is fine for now, I need to spend some time with the perf test and setup different scenarios so we can see effects of small data, large data, lots of fields vs bytes. stuff like that

PR #108, opened 2020-02-10
Acquire stuff

bbalser:


apps/service_acquire/lib/acquire/query/st.ex
```diff
@@ -16,6 +16,10 @@ defmodule Acquire.Query.ST do
     end
   end
 
+  # def envelope(geometry) do
+  #   Function.new(function: "ST_Envelope", args: [geometry])
+  # end
+
```
jdenen: ?

apps/service_acquire/lib/acquire/query/where/temporal.ex
```diff
@@ -0,0 +1,65 @@
...
+        ) :: Acquire.Queryable.t()
+  def to_queryable(_, _, "", ""), do: nil
+
+  def to_queryable(dataset_id, subset_id, after_time, before_time) do
+    with {:ok, dictionary} <- Acquire.Dictionaries.get(dataset_id, subset_id, "dictionary") do
+      Dictionary.get_by_type(dictionary, Dictionary.Type.Timestamp)
+      |> Enum.map(&Enum.join(&1, "."))
+      |> Enum.map(&to_queryable(&1, after_time, before_time))
+      |> or_clause()
+    end
```
jdenen: Let's talk about the `Acquire.Dictionaries.get` pattern that we did with WKT. We should change it here or there so it's the same.

PR #114, opened 2020-02-12
Helm

jdenen:


helm/charts/presto/Chart.yaml
```diff
@@ -0,0 +1,14 @@
+name: kubernetes-data-platform
```
jdenen: I'm going to rename this subchart.

jdenen: `presto` it is

PR #121, opened 2020-02-17
Configure Brook storage to use Redis

cameronmarsh:


config/releases.exs
```diff
@@ -59,8 +59,8 @@ config :service_gather, Gather.Application,
     ],
     handlers: [Gather.Event.Handler],
     storage: [
-      module: Brook.Storage.Ets,
-      init_arg: []
+      module: Brook.Storage.Redis,
+      init_arg: [redix_args: [host: "localhost"], namespace: "service:gather:view"]
```
jeffgrunewald: since the `releases.exs` file is intended to capture prod-ready settings, i'd suggest pulling all of these `[host: "localhost"]` out into a single variable near the top of the file (see how we reference `kafka_endpoints` and then define it as `System.get_env("REDIS_HOST") || [host: "localhost"]` to pull the value from the helm charts if they're supplied.

cameronmarsh: I just changed these to use `System.get_env("REDIS_HOST")`. Should I add `localhost` as a backup still?

jeffgrunewald: as of elixir 1.9 it's considered best practice to include a failover value for `System.get_env` calls, so i would.

apps/platform_runner/test/end_to_end_test.exs
```diff
@@ -102,19 +102,28 @@ defmodule PlatformRunner.EndToEndTest do
...
              Prestige.query(session, "select * from e2e__csv order by single_letter") do
-        assert Prestige.Result.as_maps(result) == [
-                 %{"single_letter" => "a", "number" => "1"},
-                 %{"single_letter" => "b", "number" => "2"},
-                 %{"single_letter" => "c", "number" => "3"}
-               ]
+        assert Enum.member?(Prestige.Result.as_maps(result), %{
+                 "single_letter" => "a",
+                 "number" => "1"
+               })
```
jdenen:
```elixir
assert %{"single_letter" => "a", "number" => "1"} in Prestige.Result.as_maps(result)
```

jdenen: I think this is a more readable way to assert the same thing.

config/releases.exs
```diff
@@ -41,6 +41,17 @@ kafka_endpoints =
   |> Enum.map(fn entry -> String.split(entry, ":") end)
   |> Enum.map(fn [host, port] -> {String.to_atom(host), String.to_integer(port)} end)
 
+get_redix_args = fn host, password ->
+  [host: host, password: password]
+  |> Enum.filter(fn
+    {_, nil} -> false
+    {_, ""} -> false
+    _ -> true
+  end)
```
jdenen: We decided to make authenticated Redis required.

apps/service_orchestrate/lib/orchestrate/schedule/store.ex
```diff
@@ -6,14 +6,14 @@ defmodule Orchestrate.Schedule.Store do
 
   @spec persist(Schedule.t()) :: :ok
   def persist(%Schedule{} = schedule) do
-    Brook.ViewState.merge(@collection, identifier(schedule), %{schedule: schedule})
+    Brook.ViewState.merge(@collection, identifier(schedule), %{"schedule" => schedule})
```
jdenen: Why'd you make this change? Just looking to understand.

apps/platform_runner/test/end_to_end_test.exs
cameronmarsh: I agree but we got an error on this before from the formatter

apps/service_orchestrate/lib/orchestrate/schedule/store.ex
cameronmarsh: Woops, I guess we didn't need to change this line. We had some integration tests fail when we called `Map.get` below this because calls to redis were returning string keys. I'm assuming because they are getting encoded/decoded?

I think we did this somewhere else, too. I will fix that

jdenen: No it's fine. I was just curious. Other services have a `store.ex` file with basically the same function calls. How do you feel about changing them there too? I'd like to be consistent.

apps/platform_runner/test/end_to_end_test.exs
jdenen: Try it, then run `mix format`. My guess is it was a line length issue.

apps/service_orchestrate/lib/orchestrate/schedule/store.ex
cameronmarsh: Sure, np

apps/platform_runner/test/end_to_end_test.exs
cameronmarsh: Ah, yep that was it

config/releases.exs
```diff
@@ -71,6 +76,17 @@ config :service_receive, Receive.Writer,
...
+end
+
+redix_args = get_redix_args.(System.get_env("REDIS_HOST"), System.get_env("REDIS_PASSWORD"))
```
jdenen: Remove me in favor of the `redix_args` set on line 46.

config/releases.exs
```diff
@@ -43,6 +43,11 @@ kafka_endpoints =
 
 bucket_region = System.get_env("BUCKET_REGION", "local")
 
+redix_args = [
+  host: System.get_env("REDIS_HOST") || "localhost",
```
jdenen:
```suggestion
  host: System.get_env("REDIS_HOST", "localhost"),
```

Elixir 1.9 added `System.get_env/3` so you don't have to do the `||` stuff anymore.

config/releases.exs
```diff
@@ -43,6 +43,11 @@ kafka_endpoints =
...
+  password: System.get_env("REDIS_PASSWORD")
```
jdenen: This should have a default password.

apps/platform_runner/docker-compose.yml
```diff
@@ -55,3 +55,12 @@ services:
       interval: 10s
       timeout: 30s
       retries: 10
+  redis:
```
jdenen: This Redis instance should be authenticated. I'm guessing it's not by default.

PR #124, opened 2020-02-18
Create projects for each protocol and hopefully fix the dependency graph

bbalser:


apps/extractor/test/support/test_steps.ex
```diff
@@ -4,7 +4,7 @@ defmodule Test.Steps do
 
     defimpl Extract.Step, for: CreateResponse do
       def execute(step, context) do
-        {:ok, Extract.Steps.Context.set_response(context, step.response)}
+        {:ok, Extract.Context.set_response(context, step.response)}
```
jdenen: Do you even `Ok.ok/1` bro?

PR #126, opened 2020-02-18
Extract envelope

bbalser:


apps/extract_kafka/lib/extract/kafka/subscribe.ex
```diff
@@ -36,38 +36,58 @@ defmodule Extract.Kafka.Subscribe do
...
+        "topic" => topic,
+        "partition" => partition,
+        "generation_id" => generation_id,
+        "offset" => offset
+      } =
+        messages
+        |> Enum.map(&get_in(&1, [Access.key(:meta), "kafka"]))
+        |> Enum.max_by(&Map.get(&1, "offset"))
+
+      :ok = Elsa.Group.Acknowledger.ack(connection, topic, partition, generation_id, offset)
```
jeffgrunewald: i like this a lot better than having an additional acknowledger process.

apps/extract_kafka/lib/extract/kafka/subscribe.ex
```diff
@@ -36,38 +36,58 @@ defmodule Extract.Kafka.Subscribe do
...
-          Acknowledger.cache(acknowledger, messages)
-          unwrapped_messages = Enum.map(messages, fn %{value: payload} -> payload end)
-          {unwrapped_messages, acc}
+          extract_messages =
+            Enum.map(messages, fn %{value: payload} = elsa_message ->
+              meta =
+                elsa_message
+                |> Map.from_struct()
+                |> Map.drop([:value, :timestamp, :headers, :key])
+                |> Enum.reduce(%{}, fn {k, v}, acc ->
```
jeffgrunewald: it feels like they should make it easier to convert a Map to string keys.

apps/extract_step/lib/extract/message.ex
```diff
@@ -0,0 +1,29 @@
+defmodule Extract.Message do
```
jeffgrunewald: do these ever get persisted anywhere? like in brook state or are they _totally_ for in-transit messages? i ask b/c not using the `Definition` module to allow for all of the lifecycle management of the structs could blow up in our faces if they're persisted anywhere.

apps/extract_kafka/lib/extract/kafka/subscribe.ex
```diff
@@ -36,38 +36,58 @@ defmodule Extract.Kafka.Subscribe do
 
   defimpl Extract.Step, for: __MODULE__ do
     import Extract.Context
-    alias Extract.Kafka.Subscribe.Acknowledger
+
+    @dialyzer [:no_return, :no_fail_call]
```
jdenen: You hurt my heart with this 💔

apps/extract_decode_gtfs/lib/extract/decode/gtfs.ex
```diff
@@ -11,10 +11,15 @@ defmodule Extract.Decode.Gtfs do
...
-        get_stream(context, opts)
+        data_list = get_stream(context, opts) |> Enum.to_list()
+        meta = List.last(data_list) |> Map.get(:meta)
+
+        data_list
+        |> Enum.map(&Map.get(&1, :data))
         |> Enum.join()
         |> TransitRealtime.FeedMessage.decode()
         |> Map.get(:entity)
+        |> Enum.map(&Extract.Message.new(data: &1, meta: meta))
```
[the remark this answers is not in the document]
bbalser: yeah, it was the solution Johnson and I agreed to.  The last one should always be the one with highest offset

apps/extract_step/lib/extract/message.ex
bbalser: totally in-transit

PR #128, opened 2020-02-20
More installation tweaks

jdenen:
- Make deployment work with external Postgres instance
- Add `scripts/install` script
- Configure Strimzi to watch all namespaces
- Configure hive to allow `DROP TABLE` by default

### DIFF
--- a/scripts/install
+++ b/scripts/install
@@ -1,0 +1,45 @@
+#!/usr/bin/env bash
+
+function usage {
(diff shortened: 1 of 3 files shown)


scripts/install
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env bash
```
jmorris-pillar: do you want to set -eo pipefail?

PR #129, opened 2020-02-20
Send extract_end on extract completion

cameronmarsh:


apps/service_gather/lib/gather/extraction.ex
```diff
@@ -31,6 +31,8 @@ defmodule Gather.Extraction do
 
     case extract(extract) do
       :ok ->
+        Logger.debug("#{__MODULE__}: Extraction Completed: #{inspect(extract)}")
+        Brook.Event.send(Gather.Application.instance(), "extract:end", "gather", extract)
```
jdenen:
```suggestion
        Brook.Event.send(Gather.Application.instance(), extract_end(), "gather", extract)
```

The `Events` module defines macros for all events. You'll need to import the correct event macro in this case.

PR #142, opened 2020-02-26
Receive websocket

jeffgrunewald:
Adding support for a websocket-based receive gateway.


apps/accept_websocket/lib/accept/websocket.ex
```diff
@@ -0,0 +1,40 @@
...
+    def connect(accept, opts) do
+      {
+        Accept.Websocket.Supervisor,
+        :start_link,
+        [
+          path: accept.path,
+          port: accept.port,
+          idle_timeout: accept.idle_timeout,
+          hibernate: accept.hibernate
+        ] ++ opts
```
jdenen: Is `++` what you want here? Or `Keyword.merge` with your specific configuration overriding anything that is set in `opts`?

apps/accept_websocket/test/test_helper.exs
```diff
@@ -0,0 +1 @@
+ExUnit.start(capture_log: true)
```
jdenen: 🙏

apps/testing/lib/source_websocket.ex
```diff
@@ -0,0 +1,3 @@
+defmodule SourceWebsocket do
+  use Glock.Socket
```
jdenen: > Glock.Socket

You're making this up...

jeffgrunewald: You're right, i did make it up. When i wrote the library: https://hexdocs.pm/glock

apps/accept_websocket/lib/accept/websocket.ex
jeffgrunewald: i could get behind that; they're not intended to overlap since the `opts` is where we pass in the stuff common to all receive gateways (the writer function, the batch size, and the timeout for sending messages even if a batch isn't full, not to be confused with the websocket idle_timeout for when the socket hasn't seen activity for a while.

jeffgrunewald: giving it a try; doesn't seem to make a difference so i'll allow it.

PR #152, opened 2020-03-02
get_by_type/2 workaround

jdenen:
I still don't know why we're receiving a list in any `get_by_type/2` usage, but this should work to handle it. Also logs a warning to help us track it down later.

### DIFF
--- a/apps/definition_dictionary/test/dictionary_test.exs
+++ b/apps/definition_dictionary/test/dictionary_test.exs
@@ -18,45 +18,46 @@ defmodule DictionaryTest do
       assert Dictionary.Type.String.new!(name: "name") == Dictionary.get_field(dictionary, "name")
     end
 
-    data_test "get_type returns all fields with that type" do
-      dictionary =
-        Dictionary.from_list([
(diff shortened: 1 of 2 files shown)


apps/definition_dictionary/test/dictionary_test.exs
```diff
@@ -18,45 +18,46 @@ defmodule DictionaryTest do
...
-          Dictionary.Type.List.new!(
-            name: "colors",
-            item_type: Dictionary.Type.String.new!(name: "in_list")
-          )
-        ])
+    data_test "get_by_type returns all fields with that type" do
+      dictionary = [
+        Dictionary.Type.String.new!(name: "name"),
+        Dictionary.Type.Integer.new!(name: "age"),
+        Dictionary.Type.Date.new!(name: "birthdate", format: "%Y-%m-%d"),
```
kennyatpillar: Is it worth throwing a test in here for Timestamp since that was what was giving me problems?

jdenen: I don't think so.

PR #158, opened 2020-03-03
Make deploying to both AWS and Minio easier

jdenen:
Use `./scripts/install [RELEASE] [NAMESPACE] [aws|local] [<key=value>, ...]` to deploy to AWS or a local Kubernetes.

* Add preset files
* Update install script to use preset files
* Update README to reflect new installation instructions


helm/templates/persist/deployment.yaml
```diff
@@ -46,7 +46,7 @@ spec:
               value: {{ .Values.global.objectStore.accessSecret }}
             {{- end }}
             - name: BUCKET_NAME
-              value: {{ .Values.global.objectStore.bucketName }}
+              value: {{ .Values.global.objectStore.bucketName | required "global.objectStore.bucketName must be set" }}
```
ManApart: That's neat you can use the `| required` to give nice errors like that.

helm/values.yaml
```diff
@@ -6,21 +6,21 @@ receive:
     type: ClusterIP
     port: 80
   service_tcp:
-    type: LoadBalancer
+    type: ClusterIP
```
ManApart: Since you're overriding these in the AWS version, can we make these guys NodePorts in the default? (Or override them in the helm values?)

helm/presets/aws.yaml
```diff
@@ -0,0 +1,44 @@
...
+    type: LoadBalancer
+acquire:
+  service:
+    type: LoadBalancer
+strimzi:
+  kafka:
+    resources:
+      requests:
+        cpu: 500m
+        memory: 6Gi
```
jmorris-pillar: you sure we have the space to turn this up to 6gigs?  Do we really need 6 gigs or would something like 4 be enough for a base install?

helm/presets/aws.yaml
```diff
@@ -0,0 +1,44 @@
...
+    enable: false
+    service:
+      externalAddress: ""
+    db:
+      user: ""
+      password: ""
+global:
+  objectStore:
+    region: "us-east-2"
+    accessKey: ""
```
jmorris-pillar: These are set to "false" for me locally

helm/presets/aws.yaml
```diff
@@ -0,0 +1,44 @@
...
+  zookeeper:
+    resources:
+      requests:
+        cpu: 100m
+        memory: 512Mi
+      limits:
+        cpu: 250m
+        memory: 1Gi
+redis:
+  enabled: false
```
jmorris-pillar: want to enable this for now?  Just because you put it in aws doesn't mean you want a managed service redis.

jdenen: No I want the presets to assume you're using managed services for Redis and Postgres. You can always override this at the command line.

helm/presets/aws.yaml
jdenen: `""` is false-y and better suited for what we're trying to do. `false` was me just not knowing better.

helm/presets/aws.yaml
jdenen: I've deployed it and it was working for me.

helm/values.yaml
jdenen: Done

PR #161, opened 2020-03-04
Gtfs decode as maps

cameronmarsh:


apps/extract_decode_gtfs/test/extract/decode/gtfs_test.exs
```diff
@@ -42,6 +42,7 @@ defmodule Extract.Decode.GtfsTest do
         File.read!("VehiclePositions.pb")
         |> TransitRealtime.FeedMessage.decode()
         |> Map.get(:entity)
+        |> Enum.map(&decode_struct/1)
```
jeffgrunewald: If you need this in prod code, wanna just make it a public function and then not have to redefine it in your test?

PR #170, opened 2020-03-06
Adds handling of nil values for all dictionary types

NathanielScottStevens:


apps/definition_dictionary/lib/dictionary/type/list.ex
```diff
@@ -48,6 +48,8 @@ defmodule Dictionary.Type.List do
   defimpl Dictionary.Type.Normalizer, for: __MODULE__ do
     alias Dictionary.Type.Normalizer
 
+    def normalize(_, nil), do: Ok.ok(nil)
```
ManApart: Should null list be transformed into empty list?

I'm guessing not since that could have other meanings? But it'd be nice to know lists are never null in the system. Just thinking aloud.

apps/definition_dictionary/lib/dictionary/type/map.ex
```diff
@@ -47,6 +47,8 @@ defmodule Dictionary.Type.Map do
   end
 
   defimpl Dictionary.Type.Normalizer, for: __MODULE__ do
+    def normalize(_, nil), do: Ok.ok(nil)
```
ManApart: Since we are passing null through for EVERY type, is there a way we could arrest this earlier and skip transforming nil values?

apps/definition_dictionary/lib/dictionary/type/list.ex
jdenen: I think we should keep writing is as `null` for now and look into if that's the behavior we want later. 

On smartcitiesdata, we had technical reasons to not `null` these. In hindsight, with our use of protocols, I'm not convinced we have any technical reason to do it anymore.

We _may_ want to do it for end users still, though.

apps/definition_dictionary/lib/dictionary/type/map.ex
NathanielScottStevens: Right now null is being converted to an empty string for Date and TimeStamp (I can't tell you if there's a good reason for that or not). Otherwise you could try to catch this in the `Dictionary` module. It might be nice to go ahead and keep it piecemeal anyway until the system has matured a little more.

PR #174, opened 2020-03-06
Lowercase all the things

bbalser:


apps/extract_decode_csv/mix.exs
```diff
@@ -27,7 +27,7 @@ defmodule ExtractDecodeCsv.MixProject do
       {:extract_step, in_umbrella: true},
       {:nimble_csv, "~> 0.6.0"},
       {:dialyxir, "~> 1.0.0-rc.7", only: [:dev], runtime: false},
-      {:checkov, "~> 1.0", only: [:test]},
+      {:checkov, "~> 1.0", only: [:dev, :test]},
```
jeffgrunewald: do we need checkov in dev anywhere or would it be better to change the locations where it already listed as `dev, test` to just `test`?

bbalser: we need dev because we import the formatter options.

PR #176, opened 2020-03-11
Partitions wow

bbalser:


apps/broadway_configurator/lib/broadway_configurator/off_broadway_kafka.ex
```diff
@@ -0,0 +1,37 @@
...
+             endpoints: endpoints(context),
+             connection: connection(context),
+             group_consumer: [
+               group: group(context),
+               topics: topics(context),
+               config: [
+                 begin_offset: :earliest,
+                 offset_reset_policy: offset_reset_policy(context),
+                 prefetch_count: 0,
+                 prefetch_bytes: 2_097_152
```
jeffgrunewald: Guessing this being in a macro means this isn’t a good place for a module attribute to define that magic number there? (which I’m sure is actually 2 mb but is still random looking)

apps/broadway_configurator/test/test_helper.exs
```diff
@@ -0,0 +1 @@
+ExUnit.start()
```
jeffgrunewald: capture_log: true?

apps/service_broadcast/lib/broadcast/stream/broadway.ex
```diff
@@ -9,8 +9,8 @@ defmodule Broadcast.Stream.Broadway do
 
   @app_name get_config_value(:app_name, required: true)
 
-  getter(:broadway_config, required: true)
   getter(:dlq, default: Broadcast.DLQ)
+  getter(:configuration, default: Broadcast.Stream.Broadway.Configuration)
```
jeffgrunewald: if this is specific to broadway, i actually prefer the old `broadway_config` getter name to make it clear you're dealing with _just_ broadway's config.

apps/service_persist/lib/persist/load/broadway.ex
```diff
@@ -74,29 +80,29 @@ defmodule Persist.Load.Broadway do
...
-        |> Keyword.update(:group_consumer, [], fn group_consumer ->
-          group_consumer
-          |> Keyword.put(:group, "persist-#{load.source}")
-          |> Keyword.put(:topics, [load.source])
-        end)
-
-      {module, config}
-    end)
-  end
+  # defp setup_config(load, transformer, writer) do
```
NathanielScottStevens: Delete this?

apps/service_persist/lib/persist/load/broadway.ex
```diff
@@ -74,29 +80,29 @@ defmodule Persist.Load.Broadway do
...
+  #       |> Keyword.put(:connection, :"persist_connection_#{load.source}")
+  #       |> Keyword.update(:group_consumer, [], fn group_consumer ->
+  #         group_consumer
+  #         |> Keyword.put(:group, "persist-#{load.source}")
+  #         |> Keyword.put(:topics, [load.source])
+  #       end)
+
+  #     {module, config}
+  #   end)
+  # end
```
jeffgrunewald: this still under development or can these be uncommented/deleted?

apps/service_persist/lib/persist/load/broadway.ex
bbalser: yup, good catch

PR #177, opened 2020-03-11
File cleanup

jdenen:
Updating #133 and addressing the PR feedback.


jdenen:
Still have to address @bbalser's feedback, which will require a larger change, so don't approve just yet.

PR #179, opened 2020-03-13
Ingress controller

jeffgrunewald:


.tool-versions
```diff
@@ -2,4 +2,5 @@ erlang 22.2.1
 erlang 22.2.4
 elixir 1.9.4
 elixir 1.9.4-otp-22
+helm 3.1.1
```
jdenen: I believe @jmorris-pillar found a breaking change from `3.0` to `3.1`.

helm/presets/aws.yaml
```diff
@@ -36,9 +36,23 @@ strimzi:
       limits:
         cpu: 250m
         memory: 512Mi
+ingress:
+  hosts:
+    - host: hindsight.com
```
jdenen: We should consider setting this to `""` and requiring it via a Go macro. We're doing this elsewhere.

.tool-versions
jeffgrunewald: any idea what it was? i didn't see any issues.

jdenen: You'll have to ask him. I can't remember for the life of me.

helm/presets/aws.yaml
jeffgrunewald: the `required` function in the templates?
i don't really like any of the ways to handle this regardless since it's handled as a list and to override any part of it you basically have to override the whole thing. If we want to be more authoritative in how people set up their ingress rules this gets easier, but right now you can arbitrarily set any number of virtual hostnames to respond to an arbitrary number of paths which makes setting up an open-ended values file with sane defaults tricky.

I can save on the complexity by having a single hostname respond to one path each for acquire, broadcast, and eventually receive, too. But someone else may want to have a different virtual host handle access to each service with multiple paths or no paths each (each service just goes to that host's `"/"`) and that's totally valid too.

Thoughts?

jdenen: I think it's completely fine if we want to be opinionated up front about how ingress should be set up. Anyone who wants something else can download the chart, make their change, and use that change.

So my vote is for setting the single hostname (that is configurable and required).

.tool-versions
jeffgrunewald: i'll leave it, as this seems to be resolved after discussing (issues with the `required` macro that i didn't encounter); either version should be fine.

helm/presets/aws.yaml
jeffgrunewald: Addresses; assumes a single host for the moment.

PR #180, opened 2020-03-17
Service profile

bbalser:


apps/definition_profile_update/lib/profile_update.ex
```diff
@@ -0,0 +1,36 @@
...
+
+  defstruct version: 1,
+            dataset_id: nil,
+            subset_id: nil,
+            updated: nil,
+            stats: %{}
+
+  def on_new(update) do
+    Map.put(update, :updated, NaiveDateTime.utc_now() |> NaiveDateTime.to_iso8601())
+    |> Ok.ok()
```
jdenen:
```suggestion
    Map.put(update, :updated, NaiveDateTime.utc_now() 
    |> NaiveDateTime.to_iso8601())
    |> Ok.ok()
```

You monster

apps/definition_profile_update/test/test_helper.exs
```diff
@@ -0,0 +1 @@
+ExUnit.start()
```
jeffgrunewald: capture_log: true

apps/service_profile/lib/profile.ex
```diff
@@ -0,0 +1,18 @@
+defmodule Profile do
```
jeffgrunewald: delete me

apps/service_profile/test/test_helper.exs
```diff
@@ -0,0 +1 @@
+ExUnit.start()
```
jeffgrunewald: capture_log: true

apps/service_profile/test/profile/feed/flow_test.exs
```diff
@@ -0,0 +1,109 @@
...
+    :ok
+  end
+
+  test "aggregates message over a configured period and calls to configured reducers" do
+    dataset_id = "ds1"
+    subset_id = "sb1"
+
+    reducers = [
+      Profile.Reducer.MinMax.new(path: [Access.key(:value), "count"]),
+      Profile.Reducer.LongestString.new(path: [Access.key(:value), "name"])
```
jeffgrunewald: assume these haven't been implemented or at least pushed up yet?

jeffgrunewald: or do you get these for free with Flow?

bbalser: they are in test/support

apps/service_profile/lib/profile/event/handler.ex
```diff
@@ -0,0 +1,14 @@
+defmodule Profile.Event.Handler do
+  use Brook.Event.Handler
+
+  import Events, only: [extract_start: 0, profile_update: 0]
+
+  def handle_event(%Brook.Event{type: extract_start(), data: %Extract{} = extract}) do
+    Profile.Feed.Supervisor.start_child(extract)
```
jeffgrunewald: do we handle an extract_end yet? how are we cleaning up a feed without a corollary "end event"?

bbalser: An extract_end event can happen alot with a dataset that is ran on cron.   To my knowledge we haven't sat down and determined how we will know when a dataset is no longer being ingested.

jeffgrunewald: so is the plan for this supervisor child to just sit there and wait until more data arrives?

bbalser: I believe so, isn't that the way broadcast and persist work? or has there been a change to those?

jeffgrunewald: they both handle `persist:end` events to clean up their children; nothing about `extract:end` but then they don't respond to `extract:start` to begin with, just the `broadcast/load:persist:start/end`

bbalser: extract:end doesn't make sense, but maybe a profile:end could be created, but like persist:end and broadcast:end, we don't have anyway to know when to send them.

jeffgrunewald: feels safer to have it so we don't have this one service failing to clean up after its children when we do institute end event handling. if all the other apps already handle those events that makes missing one more likely imo

bbalser: my vote would be we groom stories about how we are going to clean up datasets in an automated fashion.  persist:end and broadcast:end are coded, but I am not sure that is even useful since nothing sends them and I am not sure we want something to track which loads have been sent for which dataset.  We might want a more general dataset is done being ingested event.

apps/service_profile/lib/profile/feed/flow.ex
```diff
@@ -0,0 +1,69 @@
...
+    subset_id = Keyword.fetch!(opts, :subset_id)
+    from_specs = Keyword.fetch!(opts, :from_specs)
+    into_specs = Keyword.fetch!(opts, :into_specs) |> Enum.map(fn consumer -> {consumer, []} end)
+    reducers = Keyword.fetch!(opts, :reducers)
+
+    window = Flow.Window.periodic(window_limit(), window_unit())
+
+    {:ok, stats} = Profile.Feed.Store.get_stats(dataset_id, subset_id)
+    reducers = Enum.map(reducers, &Profile.Reducer.init(&1, stats))
+    {:ok, storage} = Storage.start_link(reducers: reducers)
```
jeffgrunewald: This name seems to imply the Store modules throughout the umbrella that refer to Brook; I think it makes it less clear with such similar names.

apps/service_profile/lib/profile/reducer/bounding_box.ex
```diff
@@ -0,0 +1,69 @@
+defmodule Profile.Reducer.BoundingBox do
+  defstruct [:latitude_path, :longitude_path, :xmin, :ymin, :xmax, :ymax]
```
jeffgrunewald: are we skipping `:zmin` and `:zmax` for now?

apps/service_profile/lib/profile/reducer/temporal_range.ex
```diff
@@ -0,0 +1,71 @@
...
+    defp first(a, b) do
+      case NaiveDateTime.diff(a, b) do
+        x when x > 0 -> b
+        _ -> a
+      end
+    end
+
+    defp last(nil, nil), do: nil
+    defp last(a, nil), do: a
+    defp last(nil, b), do: b
```
jeffgrunewald: no `safe_first` and `safe_last`?

apps/service_profile/mix.exs
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+  defp deps do
+    [
+      {:annotated_retry, in_umbrella: true},
+      {:brook, "~> 0.5"},
+      {:extractor, in_umbrella: true},
+      {:definition_dictionary, in_umbrella: true},
+      {:definition_events, in_umbrella: true},
+      {:elsa, "~> 0.12", override: true},
```
jeffgrunewald: do we still need `override: true`? i think everything should be up to date now.

apps/service_profile/test/profile/feed_test.exs
```diff
@@ -0,0 +1,63 @@
+defmodule Profile.FeedTest do
+  use ExUnit.Case
+
+  describe "determine_reducers/1" do
+    test "should init a temporal reduder for a timestamp field" do
```
jeffgrunewald: reduder, where's my timestamp?

apps/service_profile/test/profile/reducer/bounding_box_test.exs
```diff
@@ -0,0 +1,81 @@
...
+      where [
+        [:bbox, :long, :lat, :xmin, :ymin, :xmax, :ymax],
+        [[1.0, 2.0, 3.0, 4.0], 4.0, 1.0, 1.0, 1.0, 4.0, 4.0],
+        [[nil, nil, nil, nil], 2.7, 4.7, 2.7, 4.7, 2.7, 4.7]
+      ]
+    end
+  end
+
+  describe "merge/2" do
+    data_test "something", %{reducer: reducer} do
```
jeffgrunewald: something something something merges two bounding boxes

apps/service_profile/test/profile/reducer/temporal_range_test.exs
```diff
@@ -0,0 +1,100 @@
...
+      event = %{"ts" => ~N[2018-04-05 02:03:04] |> NaiveDateTime.to_iso8601()}
+
+      output = Profile.Reducer.init(reducer, stats) |> Profile.Reducer.reduce(event)
+
+      assert output.last == ~N[2020-01-01 01:01:01]
+    end
+  end
+
+  describe "merge/2" do
+    test "can merge to instances of the reducer", %{reducer: reducer} do
```
jeffgrunewald: nitpick: `two` instances

apps/service_profile/lib/profile/feed.ex
```diff
@@ -0,0 +1,98 @@
...
+      _ -> reducers
+    end
+  end
+
+  defp find_reducer(reducers, %struct{}) do
+    Enum.find(reducers, fn
+      %^struct{} -> true
+      _ -> false
+    end)
+  end
```
jdenen: I really struggled to understand the logic here. I re-read it a few times before it clicked, but I think the reason is because we don't always `add_to_reducers`. When a reducer exists, we don't add a new one, right? That was hard to decipher for me.

jdenen: I'm going to suggest this change:

```elixir
  defp find_temporal_range(reducers, dictionary, path) do
    case Enum.find(dictionary, fn %type{} -> type == Dictionary.Type.Timestamp end) do
      nil ->
        reducers

      field ->
        reducer = Profile.Reducer.TemporalRange.new(path: path ++ [field.name])

        case includes_reducer?(reducers, reducer) do
          true -> reducers
          false -> reducers ++ [reducer]
        end
    end
  end

  defp includes_reducer?(reducers, reducer) do
    case find_reducer(reducers, reducer) do
      nil -> false
      _ -> true
    end
  end

  defp find_reducer(reducers, %struct{}) do
    Enum.find(reducers, fn
      %^struct{} -> true
      _ -> false
    end)
  end
```

jdenen: IMO, by using a predicate like `includes_reducer?/2` up front, it's clearer that we're adding a reducer to the list when it doesn't already exist.

jeffgrunewald: is `includes_reducer?/2` really giving you anything in that scenario? You're basically just reversing the result of `find_reduce/2` then.

If that's the case, why not just:
```
unless find_reducer(reducers, reducer) do
  reducers ++ [reducer]
else
  reducers
end
```

jeffgrunewald: and then name `find_reducer/2` whatever you want to be clear if the name is still problematic for the reason you mentioned.

bbalser: That is fair but I would like to avoid the nested case statements

jeffgrunewald: my suggestion wouldn't use any case statements

jeffgrunewald:
```elixir
defp find_temporal_range(reducers, dictionary, path) do
  case Enum.find(dictionary, fn %type{} -> type == Dictionary.Type.Timestamp end) do
    nil ->
      reducers
    field ->
      reducer = Profile.Reducer.TemporalRange.new(path: path ++ [field.name])

     unless find_reducer(reducers, reducer) do
       reducers ++ [reducer]
     else
       reducers
     end
  end
end

defp find_reducer(reducers, %struct{}) do
  Enum.find(reducers, fn
    %^struct{} -> true
    _ -> false
  end)
end
```
I take that back; it doesn't _add_ any case statements.

apps/service_profile/lib/profile/reducer/bounding_box.ex
bbalser: according to johnson, yes

apps/service_profile/lib/profile/reducer/temporal_range.ex
bbalser: I can do that.  I only named min/max safe_min/safe_max because of the collision with Kernel.

apps/service_profile/lib/profile/feed/flow.ex
bbalser: I changed it to Profile.Feed.Flow.State.  Does that sound better?

apps/service_profile/lib/profile/feed.ex
bbalser: I took a crack at rewriting it from scratch, let me know if it is better or worse.

apps/service_profile/lib/profile/feed/flow.ex
jeffgrunewald: Yup

apps/service_profile/lib/profile/feed.ex
```diff
@@ -52,47 +53,42 @@ defmodule Profile.Feed do
     determine_reducers(dictionaries, reducers)
   end
 
-  defp find_bounding_box(reducers, dictionary, path) do
+  defp maybe_add_reducers(reducers, reducer, dictionary, path) do
+    unless includes_reducer?(reducers, reducer) do
+      reducers ++ create_reducers(reducer, dictionary, path)
+    else
+      reducers
+    end
```
jdenen:
```suggestion
    reducers ++ create_reducers(reducer, dictionary, path)
```

This is all you need, isn't it? If a reducer exists, then `create_reducers` will return `[]`.

bbalser: I don't think, so with this change, 2 timestamps in the dictionary would result in 2 temporal_range reducers

jeffgrunewald: that sounds right to me.

jdenen: Yeah sorry you're right. I'm trying to read code in meetings and didn't fully work through that logic.

PR #182, opened 2020-03-18
Delete dataset (kill -9 method)

kennyatpillar:
Brutally remove all traces of a dataset. Any kind of graceful cleanup is out of scope and will be addressed in a later card.


apps/platform_runner/test/end_to_end_test.exs
```diff
@@ -289,6 +289,20 @@ defmodule PlatformRunner.EndToEndTest do
...
+    test "deleted" do
+      delete =
+        Delete.new!(id: "e2e-json-delete-1", dataset_id: "e2e-json-ds", subset_id: "json-subset")
+
+      Orchestrate.Application.instance()
+      |> Events.send_definition_delete("e2e", delete)
+
+      assert_async sleep: 500, max_tries: 10 do
+        # We error out because that data / definition no longer exist
+        assert {:ok, %{"errors" => %{"detail" => "Internal Server Error"}}} =
```
NathanielScottStevens: This feels like a weird way to assert that the dataset has been removed. But maybe we just roll with it for now until Acquire gives a dataset missing error or something of the sort?

ManApart: Yeah, that's just an artefact of the

apps/service_gather/lib/gather/extraction/store.ex
```diff
@@ -21,6 +21,19 @@ defmodule Gather.Extraction.Store do
     end
   end
 
+  @spec mark_done(Extract.t()) :: :ok
+  def mark_done(%Extract{} = extract) do
+    Brook.ViewState.merge(@collection, identifier(extract), %{"done" => true})
```
NathanielScottStevens: Why marking it done vs removing it from the collection?

apps/service_orchestrate/lib/orchestrate/schedule/store.ex
```diff
@@ -26,6 +26,19 @@ defmodule Orchestrate.Schedule.Store do
     end
   end
 
+  @spec mark_done(Schedule.t()) :: :ok
+  def mark_done(%Schedule{} = schedule) do
+    Brook.ViewState.merge(@collection, identifier(schedule), %{"done" => true})
```
NathanielScottStevens: Same question here

apps/service_orchestrate/test/orchestrate/event/handler_test.exs
```diff
@@ -172,4 +172,37 @@ defmodule Orchestrate.Event.HandlerTest do
...
+    delete = %Delete{id: "123", dataset_id: schedule.dataset_id, subset_id: schedule.subset_id}
+
+    Brook.Test.send(@instance, definition_delete(), "testing", delete)
+
+    assert_async do
+      assert nil ==
+               Orchestrate.Scheduler.find_job(:"#{schedule.dataset_id}__#{schedule.subset_id}")
+    end
+
+    # We should be testing this but are not due to JSR
```
NathanielScottStevens: Why is the JSR not there? (Also, is it worth leaving the comment hanging around?)

apps/definition_events/lib/events.ex
```diff
@@ -12,6 +12,7 @@ defmodule Events do
     {"load:persist:end", Load.Persist},
     {"schedule:start", Schedule},
     {"schedule:end", Schedule},
+    {"definition:delete", Delete},
```
jeffgrunewald: definition is just a naming convention for the apps themselves that define data structures; this should just be "delete" or maybe "dataset:delete".

apps/definition_delete/test/test_helper.exs
```diff
@@ -0,0 +1 @@
+ExUnit.start()
```
jeffgrunewald:
```suggestion
ExUnit.start(capture_log: true)
```

apps/service_gather/lib/gather/extraction/store.ex
ManApart: Per Conversation with Johnson. When you go to delete a dataset, the app often needs information in the dataset-state to know for instance, what kafka topic or DB table to delete. If we delete the state at a premature point, we no longer have that needed state.

apps/service_orchestrate/lib/orchestrate/schedule/store.ex
ManApart: Same answer. :)

apps/definition_events/lib/events.ex
ManApart: I'm cool with "dataset:delete"

apps/definition_delete/test/test_helper.exs
ManApart: Good call

apps/service_orchestrate/test/orchestrate/event/handler_test.exs
ManApart: I knew someone was going to ask, and now I've forgotten everything but the pain. Let me see if I can remember why.

apps/service_broadcast/lib/broadcast/event/handler.ex
```diff
@@ -15,20 +24,46 @@ defmodule Broadcast.Event.Handler do
...
+        nil
+
+      load ->
+        terminate_stream(load)
+        if Elsa.topic?(endpoints(), load.source), do: Elsa.delete_topic(endpoints(), load.source)
+        Logger.debug("Deleted kafka topic")
+    end
+
+    Broadcast.Transformations.delete(delete)
+    Broadcast.Stream.Store.delete(delete.dataset_id, delete.subset_id)
```
jdenen: Add an INFO message after all deletions are complete.

apps/service_gather/lib/gather/event/handler.ex
```diff
@@ -13,6 +16,26 @@ defmodule Gather.Event.Handler do
...
+        nil
+
+      extract ->
+        if Elsa.topic?(endpoints(), extract.destination),
+          do: Elsa.delete_topic(endpoints(), extract.destination)
+
+        Logger.debug("Deleted Extract destination topic")
+    end
+
+    Extraction.Store.delete(delete.dataset_id, delete.subset_id)
```
jdenen: Same deal here. Add an INFO message.

apps/service_orchestrate/lib/orchestrate/event/handler.ex
```diff
@@ -34,7 +35,17 @@ defmodule Orchestrate.Event.Handler do
...
+  end
+
+  def handle_event(%Brook.Event{type: dataset_delete(), data: %Delete{} = delete}) do
+    Logger.debug(fn ->
+      "#{__MODULE__}: Received event #{dataset_delete()}: #{inspect(delete)}"
+    end)
+
+    Orchestrate.Scheduler.delete_job(:"#{identifier(delete)}")
+    Orchestrate.Scheduler.delete_job(:"#{identifier(delete)}_compaction")
+    Orchestrate.Schedule.Store.delete(delete.dataset_id, delete.subset_id)
```
jdenen: INFO

apps/service_persist/lib/persist/event/handler.ex
```diff
@@ -55,8 +57,31 @@ defmodule Persist.Event.Handler do
...
+      load ->
+        Persist.Compact.Supervisor.terminate_child(load)
+        Persist.Load.Supervisor.terminate_child(load)
+        Persist.TableManager.delete(load.destination)
+        Persist.TableManager.delete("#{load.destination}__staging")
+        if Elsa.topic?(endpoints(), load.source), do: Elsa.delete_topic(endpoints(), load.source)
+        Logger.debug("Deleted existing state")
+    end
+
+    Persist.Load.Store.delete(delete.dataset_id, delete.subset_id)
```
jdenen: INFO

apps/service_receive/lib/receive/event/handler.ex
```diff
@@ -13,16 +16,45 @@ defmodule Receive.Event.Handler do
...
+      accept ->
+        terminate_manager(accept)
+
+        if Elsa.topic?(endpoints(), accept.destination),
+          do: Elsa.delete_topic(endpoints(), accept.destination)
+
+        Logger.debug("Deleted Supervisor and Topic")
+    end
+
+    Receive.Accept.Store.delete(delete.dataset_id, delete.subset_id)
```
jdenen: INFO

apps/service_acquire/lib/acquire/event/handler.ex
```diff
@@ -10,4 +11,13 @@ defmodule Acquire.Event.Handler do
...
     Acquire.Dictionaries.persist(persist)
   end
+
+  def handle_event(%Brook.Event{type: dataset_delete(), data: %Delete{} = delete}) do
+    Logger.debug(fn ->
+      "#{__MODULE__}: Received event #{dataset_delete()}: #{inspect(delete)}"
+    end)
+
+    Acquire.Dictionaries.delete(delete)
+    Logger.debug("Deleted Dictionaries")
```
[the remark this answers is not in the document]
jeffgrunewald: agree on 1; won't you get the two ids from inspecting the delete struct? or do you mean just make them easier to see in the log message?

apps/service_broadcast/lib/broadcast/event/handler.ex
```diff
@@ -15,20 +24,46 @@ defmodule Broadcast.Event.Handler do
...
+  end
+
+  def handle_event(%Brook.Event{type: dataset_delete(), data: %Delete{} = delete}) do
+    Logger.debug(fn ->
+      "#{__MODULE__}: Received event #{dataset_delete()}: #{inspect(delete)}"
+    end)
+
+    case Broadcast.Stream.Store.get!(delete.dataset_id, delete.subset_id) do
+      nil ->
+        Logger.debug("No existing state to delete")
```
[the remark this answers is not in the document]
jeffgrunewald: we should probably also standardize on all the log messages taking the desired statement wrapped in an anonymous function, too. You have them elsewhere in the PR, such as on line 35 above.

apps/service_broadcast/lib/broadcast/event/handler.ex
```diff
@@ -15,20 +24,46 @@ defmodule Broadcast.Event.Handler do
...
+        nil
+
+      load ->
+        terminate_stream(load)
+        if Elsa.topic?(endpoints(), load.source), do: Elsa.delete_topic(endpoints(), load.source)
+        Logger.debug("Deleted kafka topic")
```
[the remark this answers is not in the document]
jeffgrunewald: i would just try and delete the topic and handle the {:error, msg} in the event another app has already done the job; saves us making twice as many calls to kafka

apps/service_broadcast/lib/broadcast/init.ex
```diff
@@ -5,6 +5,8 @@ defmodule Broadcast.Init do
 
   def on_start(state) do
     Broadcast.Stream.Store.get_all!()
+    |> Enum.reject(&is_nil/1)
```
jeffgrunewald: why would we be getting `nil`s back? is this a side effect of confusion when we're querying expecting an atom key and all maps in the view state are going to have string keys?

apps/service_broadcast/lib/broadcast/init.ex
```diff
@@ -5,6 +5,8 @@ defmodule Broadcast.Init do
...
+    |> Enum.reject(&Broadcast.Stream.Store.done?(&1))
```
jeffgrunewald: would it make more sense to move "done" things to a different collection so they're still able to be referenced but not clogging up the works for regular operations of the app?

apps/service_broadcast/lib/broadcast/stream/broadway.ex
```diff
@@ -21,7 +21,7 @@ defmodule Broadcast.Stream.Broadway do
     %Load.Broadcast{} = load = Keyword.fetch!(init_arg, :load)
     Logger.debug(fn -> "#{__MODULE__}: Starting for #{inspect(load)}" end)
 
-    with {:ok, transformer} <- create_transformer(load.dataset_id),
+    with {:ok, transformer} <- create_transformer(load.dataset_id, load.subset_id),
```
jeffgrunewald: good catch on this change.

apps/service_broadcast/lib/broadcast/stream/store.ex
```diff
@@ -9,7 +9,20 @@ defmodule Broadcast.Stream.Store do
...
+
+  @spec done?(Load.Broadcast.t()) :: boolean
+  def done?(%Load.Broadcast{} = load) do
+    case Brook.get!(@instance, @collection, identifier(load)) do
+      nil -> false
+      map -> Map.get(map, "done", false)
+    end
+  end
+
+  @spec get!(dataset_id :: String.t(), subset_id :: String.t()) :: Load.Broadcast.t() | nil
```
jeffgrunewald: same question as before? does it make sense to let these keep piling up if they're only there for infrequent references or would we want to create a second collection and migrate completed records to that one?

apps/service_broadcast/lib/broadcast/init.ex
ManApart: I believe so. It could also come from deleting the inner map (event/struct) but leaving transformations or something else laying around. In the end it seemed better to be defensive and save ourselves from a future-footgun.

apps/service_broadcast/lib/broadcast/init.ex
ManApart: I'm for that if we get to a point where we see it affect performance.

apps/service_gather/lib/gather/event/handler.ex
```diff
@@ -13,6 +16,26 @@ defmodule Gather.Event.Handler do
...
+    case Extraction.Store.get!(delete.dataset_id, delete.subset_id) do
+      nil ->
+        Logger.debug("No state to delete")
+        nil
+
+      extract ->
+        if Elsa.topic?(endpoints(), extract.destination),
+          do: Elsa.delete_topic(endpoints(), extract.destination)
+
+        Logger.debug("Deleted Extract destination topic")
```
jeffgrunewald: similar comments re: log messages wrapped in anonymous functions and just arbitrarily deleting the topic.

apps/service_broadcast/lib/broadcast/stream/store.ex
ManApart: It doesn't feel like the JSR is there to do that right now, but if the team prefers that we could go back and change it everywhere. I think it's a good call out, but I think it could sacrifice simplicity for performance, and I'd like performance to drive that.

apps/service_broadcast/lib/broadcast/event/handler.ex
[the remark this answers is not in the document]
ManApart: I remember people talking about wrapping log statements in lambdas. I'd like to better understand _why_ it's more performant. But either way it makes sense that we should update them to use the lambda.

apps/service_broadcast/lib/broadcast/event/handler.ex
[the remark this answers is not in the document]
ManApart: That makes sense.

apps/service_broadcast/lib/broadcast/init.ex
jeffgrunewald: feels like if we're persisting `nil`s into an app's view state that _is_ a footgun

apps/service_broadcast/lib/broadcast/stream/store.ex
jeffgrunewald: until we have a paying client with a timeline who's money we're burning through, i feel like the "J" is "the best possible open source project that someone might find valuable". To me, the whole concept of JSR is weighing the pros and cons of an action with respect to a budget and a timeline, neither of which are really applicable here so now's the time to be thinking about architectural choices that might be painful later.

jeffgrunewald: if this (or anything) is actually a bad idea, or it's not worth doing within a specific story that's a different matter, but for as long as we're not being paid by a client on a timeline I want to push back on the idea of "action not being worth the effort" especially when it relates to an open source project that, independently of any particular use we may find for it, has no real timeline or budgetary constraints.

apps/service_persist/lib/persist/table_manager.ex
```diff
@@ -22,6 +23,10 @@ defmodule Persist.TableManager do
   def copy(from_table, to_table) do
     impl().copy(from_table, to_table)
   end
+
+  def delete(table) do
```
jeffgrunewald: do these need `@spec` or `@impl` annotations since they're public functions?

apps/service_acquire/lib/acquire/event/handler.ex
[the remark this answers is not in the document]
jdenen: I just mean include them in the log message. I don't particularly care how, whether it's inspecting the struct or interpolating them into the string.

kennyatpillar:
@jeffgrunewald on your command about "why are we marking things done in a PR meant to delete the dataset" Once we implemented this, we found that in some events like "schedule:end" some state was being deleted. So we're marking things done when they are done, and deleting it in the delete event. If there's a time we are marking as done when we should have deleted, let us know.

apps/service_broadcast/lib/broadcast/init.ex
ManApart: That's orthogonal to whether we mark done or do another collection. We should NOT be persisting nils, but the code was already persisting nils when we grabbed this card, due to the atoms / strings thing. We're rejecting nils here so that if other code is written like the atom keys thing later, we're not back into a broken state.

apps/service_persist/lib/persist/table_manager.ex
ManApart: Maybe and I'm just not aware. I don't see def/impl on any of the sibling functions in this module, and dialyzer passes. 🤷‍♂

ManApart:
To Jeff's comment: We are COMPLETELY deleting the dataset from viewstate on reception of a delete event. We're just NOT deleting event state if say, a load has finished. This allows us to completely clean up a dataset without having to force the dataset event to know all implementation of every app. 

Separate collections is another way we could solve this issue. I believe that that would create more complexity in the app (two gets or a third param etc) and is not worth the perceived performance benefit at this time. I'm fine with us doing two collections if we decide it's worth it.

I think we should always be asking about JSR, and if we're not that's a red flag to me. That being said, the JSR of JSR is probably not best discussed in github comments.

To Johnson. Sounds good, I'd like to touch base with you on monday then.

apps/service_broadcast/lib/broadcast/event/handler.ex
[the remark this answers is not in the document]
bbalser: it is more performant because the log string is not computed unless debug is turned on.  In most cases we are adding dynamic values to the log message and if you don't use an anonymous function it then computes that string and passes it the function which throws it away.  HTH

PR #183, opened 2020-03-19
Updated to pull secrets automatically from secretsmanager for aws dep…

jmorris-pillar:
…loys

### DIFF
--- a/scripts/install
+++ b/scripts/install
@@ -8,8 +8,8 @@ function usage {
     echo "USAGE: ./scripts/install [RELEASE_NAME] [NAMESPACE] local [<key=value>, ...]"
     echo
     echo "EXAMPLE"
-    echo "    ./scripts/install foo bar local dev abc=123 x.y.z=42"
-    echo "    ./scripts/install abc default aws dev image.tag=latest"
+    echo "    ./scripts/install foo bar local dev image.tag=development abc=123 x.y.z=42"
+    echo "    ./scripts/install abc default aws dev image.tag=latest ingress.host=mydomain.com"
     echo
     exit 0
 }
@@ -29,8 +29,6 @@ function check_strimzi {
 }
 
 function deps {
-    local -r added_p=$(helm repo list | grep -s google)
-
     info "Installing Helm dependencies"
     helm repo add google https://kubernetes-charts.storage.googleapis.com
     (cd ./helm; helm dependency build)
--- a/scripts/install
+++ b/scripts/install
@@ -1,9 +1,11 @@
 #!/usr/bin/env bash
 
-set -xeo pipefail
+set -eo pipefail
 
 function usage {
-    echo "USAGE: ./scripts/install [RELEASE_NAME] [NAMESPACE] [aws|local] [environmentName] [<key=value>, ...]"
+    echo "USAGE: ./scripts/install [RELEASE_NAME] [NAMESPACE] aws [environmentName] [<key=value>, ...]"
+    echo
+    echo "USAGE: ./scripts/install [RELEASE_NAME] [NAMESPACE] local [<key=value>, ...]"
     echo
     echo "EXAMPLE"
     echo "    ./scripts/install foo bar local dev abc=123 x.y.z=42"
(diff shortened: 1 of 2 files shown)


scripts/install
```diff
@@ -29,11 +31,9 @@ function check_strimzi {
 function deps {
     local -r added_p=$(helm repo list | grep -s google)
```
jeffgrunewald: looks like we can remove this local variable now that we're _definitely_ installing the google remote chart repo for ingress et al.

scripts/install
```diff
@@ -52,10 +52,28 @@ function strimzi {
...
     local -r location="${3:?Location required.}"
-    shift 3
+    local -r environment="${4:?Environment required.}"
+    shift 4
+
+    local -r username=$(aws secretsmanager get-secret-value --secret MetastoreInfo-${environment} | jq -rc .SecretString | jq .username | tr -d "\n\"")
+    local -r password=$(aws secretsmanager get-secret-value --secret MetastoreInfo-${environment} | jq -rc .SecretString | jq .password | tr -d "\n\"")
+    local -r postgresEndpointAddress=$(aws secretsmanager get-secret-value --secret MetastoreInfo-${environment} | jq -rc .SecretString | jq .endpoint | tr -d "\n\"")
+    local -r bucketName=$(aws secretsmanager get-secret-value --secret MetastoreInfo-${environment} | jq -rc .SecretString | jq .bucket | tr -d "\n\"")
+    local -r redisEndpointAddress=$(aws secretsmanager get-secret-value --secret RedisInfo-${environment} | jq -rc .SecretString | jq .endpoint | tr -d "\n\"")
```
jeffgrunewald: I'm okay with this for the moment but I don't think long-term we want to couple ourselves to any specific implementations of infrastructure deployments, especially in the generic project repo. I think this makes _slightly_ more sense in the inhindsight/hindsight-aws repo but i think it would be preferable to assume a standard set of flags that get passed to our script or have the script take a config file and it's an exercise left to the reader how they generate those flags or that config file

@jdenen thoughts?

jmorris-pillar: This is probably fair, although making things and exercise for the reader means they're an exercise for us and I want this thing to be easily deployed without having to jump through hoops.  I am cool with these implementation details /for us/ living somewhere else though and would be behind hindsight proper not leaking those.  Yesterday johnson was cool with not making that move yet but id be interested to see if he changes his mind, im ambivelent in the short term.

scripts/install
jmorris-pillar: good call

scripts/install
```diff
@@ -52,10 +50,28 @@ function strimzi {
...
+
+function aws_deploy {
     local -r release="${1:?Release name required.}"
     local -r ns="${2:?Namespace required.}"
     local -r location="${3:?Location required.}"
-    shift 3
+    local -r environment="${4:?Environment required.}"
+    shift 4
+
+    local -r username=$(aws secretsmanager get-secret-value --secret MetastoreInfo-${environment} | jq -rc .SecretString | jq .username | tr -d "\n\"")
```
jdenen:
```bash
function get_aws_secret {
    local -r secret="${1:?Secret required.}"
    local -r key="${2:?JSON key required.}"
    aws secretsmanager get-secret-value --secret "${secret}" | jq -rc .SecretString | jq ".${key}" | tr -d "\n\""
}

local -r password=$(get_aws_secret "MetastoreInfo-${environment}" password)
local -r postgresEndpointAddress=$(get_aws_secret "MetastoreInfo-${environment}" endpoint)
# etc
```

Feels like a place to DRY up.

scripts/install
jdenen: Punt on it for now.

This script isn't required for using Hindsight, so I don't see it as a big deal to tie ourselves to AWS directly here. If you're deploying to Azure/GCP/Digitial Ocean, don't use the script.

jdenen: Long term, I agree that we'll want to move away from this. But JSR is low for that line of work at the moment.

PR #185, opened 2020-03-23
adding prometheus operator chart

jeffgrunewald:
adds the prometheus operator helm chart to the hindsight chart deployment as a required dependency.

the chart dependency _does_ include a toggle but this is intended to allow for using an existing prometheus or other monitoring stack in the environment if you already have one. Since prometheus does not have a cloud-managed offering, the toggle is in the default values.yaml file and it is assumed to be enabled by default.

because none of the monitoring stack components have app root paths and the `nginx rewrite-target` annotation isn't working for these, they all have a separate virtual host annotation that point to the same loadbalancer (the same as the hindsight app itself).

### DIFF
--- a/helm/presets/aws.yaml
+++ b/helm/presets/aws.yaml
@@ -45,6 +45,18 @@ nginx-ingress:
       annotations:
         service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout: "3600"
       type: LoadBalancer
+prometheus-operator:
+  prometheus:
+    prometheusSpec:
+      storageSpec:
+        volumeClaimTemplate:
+          spec:
+            storageClassName: gp2
+            accessModes: ["ReadWriteOnce"]
+            resources:
+              requests:
+                storage: 100Gi
+          selector: {}
 presto:
   postgres:
     enable: false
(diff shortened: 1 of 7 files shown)


helm/presets/aws.yaml
```diff
@@ -45,6 +45,18 @@ nginx-ingress:
...
+  prometheus:
+    prometheusSpec:
+      storageSpec:
+        volumeClaimTemplate:
+          spec:
+            storageClassName: gp2
+            accessModes: ["ReadWriteOnce"]
+            resources:
+              requests:
+                storage: 100Gi
```
jeffgrunewald: This may need to grow but it's based on the amount set as the default in the smartcolumbusos chart.

helm/values.yaml
```diff
@@ -35,6 +35,22 @@ nginx-ingress:
       enabled: true
     service:
       type: LoadBalancer
+prometheus-operator:
+  enabled: true
```
jdenen: I think the local presets file should toggle this off by default, as you probably don't want prometheus on Minikube.

jeffgrunewald: if you don't think it would be a useful for gathering relative performance changes in a local environment i can get behind that. it doesn't really add much (not like the elasticsearch cluster for logging) but if it's not useful then it should get toggled off.

PR #188, opened 2020-03-26
Refactor to source/destination protocols

jdenen:


apps/service_profile/test/profile/event/handler_test.exs
```diff
@@ -21,7 +21,7 @@ defmodule Profile.Event.HandlerTest do
           subset_id: "sb1",
           steps: [],
           dictionary: [],
-          destination: "topic"
+          destination: Destination.Fake.new!()
```
jeffgrunewald: I thought this said "Fake news" at first. What a time to be alive.

apps/service_persist/test/persist_test.exs
```diff
@@ -159,12 +147,8 @@ defmodule PersistTest do
...
         dataset_id: "ds1",
         subset_id: "example",
-        source: "topic-example",
-        destination: "ds1_example",
-        schema: [
-          %Dictionary.Type.String{name: "name"},
-          %Dictionary.Type.Integer{name: "age"}
-        ]
+        source: Source.Fake.new(),
+        destination: "ds1_example"
```
jeffgrunewald: Don't both of these need to be the new style of protocol struct?

apps/service_persist/test/persist_test.exs
```diff
@@ -124,9 +113,8 @@ defmodule PersistTest do
         id: "persist-1",
         dataset_id: "ds1",
         subset_id: "example",
-        source: "topic-example",
-        destination: "ds1_example",
-        schema: []
+        source: Source.Fake.new(),
+        destination: "ds1_example"
```
jeffgrunewald: Don't both of these need to be the new type of protocol struct?

apps/protocol_destination/mix.exs
```diff
@@ -0,0 +1,29 @@
+defmodule ProtocolDefinition.MixProject do
```
jeffgrunewald: I know it doesn't matter, but this is bad and you should feel bad.

apps/definition_kafka/lib/kafka/topic.ex
```diff
@@ -0,0 +1,64 @@
...
+  def on_new(struct) do
+    endpoints =
+      struct.endpoints
+      |> List.wrap()
+      |> Enum.map(&fix_endpoint/1)
+
+    struct
+    |> Map.put(:endpoints, endpoints)
+    |> Map.update!(:partitioner, &fix_partitioner/1)
+    |> Ok.ok()
```
jeffgrunewald: Why not handle fix_endpoints and fix_partitioner in the same way? can't both be "fixed" via the `Map.update!` method and include the list wrapping as part of the fix endpoint function as needed?

apps/definition_kafka/lib/kafka/topic/destination.ex
```diff
@@ -0,0 +1,155 @@
...
+  use Properties, otp_app: :definition_kafka
+  require Logger
+
+  getter(:dlq, default: Dlq)
+
+  @spec start_link(Destination.t(), Destination.init_opts()) ::
+          {:ok, Destination.t()} | {:error, term}
+  def start_link(topic, opts) do
+    GenServer.start_link(__MODULE__, {topic, opts})
+    |> Ok.map(&%{topic | pid: &1})
```
jeffgrunewald: is the `{:ok, pid}` still returned to the caller here or is it swallowed by merging it into the topic state?

apps/definition_kafka/lib/kafka/topic/destination.ex
```diff
@@ -0,0 +1,155 @@
...
+      end)
+
+    with :ok <- write(topic, Enum.reverse(encoded.ok)) do
+      GenServer.cast(topic.pid, {:dlq, Enum.reverse(encoded.error)})
+      :ok
+    end
+  end
+
+  def write(topic, messages) do
+    with :ok <- GenServer.call(topic.pid, {:write, topic, messages}),
```
jeffgrunewald: if the topic is a process now, why do we need to send it it's own struct definition?

apps/definition_kafka/lib/kafka/topic/destination.ex
```diff
@@ -0,0 +1,155 @@
...
+    GenServer.call(topic.pid, :stop)
+  end
+
+  @spec delete(Destination.t()) :: :ok | {:error, term}
+  def delete(topic) do
+    with {:error, reason} <- Elsa.delete_topic(topic.endpoints, topic.name),
+         log_reason <- inspect(reason) do
+      Logger.warn(fn -> "Topic '#{topic.name}' failed to delete: #{log_reason}" end)
+      Ok.error(reason)
+    end
```
jeffgrunewald: don't you need an `else` or a `rescue/catch` here for when the topic succeeds being deleted?

apps/definition_kafka/lib/kafka/topic/destination.ex
```diff
@@ -0,0 +1,155 @@
...
+  def init({topic, opts}) do
+    Process.flag(:trap_exit, true)
+    state = Map.new(opts) |> Map.put(:connection, connection_name())
+    {:ok, state, {:continue, {:init, topic}}}
+  end
+
+  @impl GenServer
+  def handle_continue({:init, topic}, state) do
+    with opts <- Map.from_struct(topic) |> Enum.into([]),
+         :ok <- Elsa.create_topic(topic.endpoints, topic.name, opts),
```
jeffgrunewald: does this need to be wrapped in a conditional check or an exponential backoff retry since we've had issues with every topic suddenly wanting to recreate itself when we have an app come online and start re-initing all of the running topics it had before it went down?

apps/definition_kafka/lib/kafka/topic/destination.ex
```diff
@@ -0,0 +1,155 @@
...
+         {:ok, pid} <- start_producer(topic, state.connection) do
+      {:noreply, Map.put(state, :elsa_pid, pid)}
+    end
+  end
+
+  @impl GenServer
+  def handle_call({:write, topic, messages}, _from, state) do
+    with opts <- Map.from_struct(topic) |> Enum.into([]),
```
jeffgrunewald: if you need to do this on `:continue` and on every call to `write` is it better to just convert the struct to a list in the continue and store _that_ in state?

apps/definition_kafka/lib/kafka/topic/destination.ex
```diff
@@ -0,0 +1,155 @@
...
+         :ok <- Elsa.produce(state.connection, topic.name, messages, opts) do
+      {:reply, :ok, state}
+    else
+      error ->
+        {:reply, error, state}
+    end
+  end
+
+  def handle_call(:stop, _from, state) do
+    Logger.info(fn -> "#{__MODULE__}: Terminating by request" end)
```
jeffgrunewald: this could probably use a reference to the topic/pid in question.

apps/definition_kafka/lib/kafka/topic/destination.ex
```diff
@@ -0,0 +1,155 @@
...
+    {:stop, :normal, :ok, state}
+  end
+
+  @impl GenServer
+  def handle_cast({:dlq, []}, state) do
+    {:noreply, state}
+  end
+
+  def handle_cast({:dlq, messages}, state) do
+    Logger.debug(fn -> "#{__MODULE__}: Writing #{Enum.count(messages)} messages to DLQ" end)
```
jeffgrunewald: which dlq? is there a single dlq here or one per standard topic?

apps/definition_kafka/lib/kafka/topic/source.ex
```diff
@@ -0,0 +1,99 @@
+defmodule Kafka.Topic.Source do
+  use GenServer, shutdown: 30_000
+  use Annotated.Retry
+  require Logger
+
+  def start_link(t, %Source.Context{} = context) do
```
jeffgrunewald: can these be more consistent between source and destination? `t` feels like a generic and confusing in light of the strong shift toward protocols.

apps/definition_kafka/lib/kafka/topic/source.ex
```diff
@@ -0,0 +1,99 @@
...
+          group: "group-#{state.context.app_name}-#{state.t.name}",
+          topics: [state.t.name],
+          handler: Kafka.Topic.Source.Handler,
+          handler_init_args: state.context,
+          config: [
+            begin_offset: :earliest,
+            offset_reset_policy: offset_reset_policy || :reset_to_earliest,
+            prefetch_count: 0,
+            prefetch_bytes: 2_097_152
+          ]
```
jeffgrunewald: this feels like it should be a config or at least a separate function.

apps/service_persist/test/persist_test.exs
jdenen: Not yet. Kafka is our only destination so far. This is a presto destination.

apps/service_persist/test/persist_test.exs
jdenen: Same deal here. Not yet.

apps/protocol_destination/mix.exs
jdenen: I write `definition` instead of `destination` all the damn time. Not surprised I missed one. Will fix.

apps/definition_kafka/lib/kafka/topic.ex
jdenen: This is code largely copied over from the Kafka writer modules. I'll shore it up later.

apps/definition_kafka/lib/kafka/topic/destination.ex
jdenen: `Destination.start_link/2` calls for `{:ok, Destination.t()}` to be returned. So the pid is wrapped in the struct being returned. @bbalser may be changing that soon.

apps/definition_kafka/lib/kafka/topic/destination.ex
jdenen: I'm going to unmake this a process soon.

apps/definition_kafka/lib/kafka/topic/destination.ex
jdenen: No. Without an `else` clause, `with` will return anything that didn't match. So if the delete returns `:ok`, then the whole thing will return `:ok`.

apps/definition_kafka/lib/kafka/topic/destination.ex
jdenen: Probably, but I'm going to un-process this whole deal soon so it's all going to change anyway.

apps/definition_kafka/lib/kafka/topic/destination.ex
jdenen: Again, un-processing this process. Please hold.

apps/definition_kafka/lib/kafka/topic/destination.ex
jdenen: There is now one running `Dlq` application.

apps/definition_kafka/lib/kafka/topic/destination.ex
jeffgrunewald: if this is going to not be a process soon that makes sense; from what i can tell this should still work from a supervision strategy since it's linked to the caller and that should be under a dynamic supervisor but it just felt like a trap only burying that pid in another process's state.

ManApart:
Holy moly this is a lot of files changed.

apps/broadway_configurator/lib/broadway_configurator.ex
```diff
@@ -1,44 +0,0 @@
-defmodule BroadwayConfigurator do
```
ManApart: I think Jeff mentioned this earlier; I'd also be curious as to why we no longer need the broadway_configurator.

jeffgrunewald: I don’t even see broadway period

apps/decoder_gtfs/lib/decoder/gtfs.ex
```diff
@@ -0,0 +1,40 @@
+defmodule Decoder.Gtfs do
```
ManApart: These 'wholly new' decoder modules and tests are actually just renames, right?

apps/definition/lib/definition/schema/type.ex
```diff
@@ -48,4 +48,14 @@ defmodule Definition.Schema.Type do
   def access_path do
     one_of([spec(is_binary()), coll_of(spec(is_binary()))])
   end
+
+  @spec impl_of(module) :: spec
+  def impl_of(protocol) do
```
ManApart: If I understand this correctly (and I'm not sure I do), I could see this being really handy.

apps/definition_deadletter/.gitignore
```diff
@@ -20,5 +20,5 @@ erl_crash.dump
 *.ez
 
 # Ignore package tarball (built via "mix hex.build").
-broadway_configurator-*.tar
+definition_deadletter-*.tar
```
ManApart: Is this file an actual rename or is git just confused by you guys removing one app and adding another?

apps/definition_extract/lib/extract.ex
```diff
@@ -8,22 +8,20 @@ defmodule Extract do
...
           subset_id: String.t(),
-          destination: String.t(),
-          steps: [Extract.Step.t()],
-          dictionary: Dictionary.t(),
-          message_key: list,
-          config: map
+          source: Source.t(),
+          decoder: Decoder.t(),
+          destination: Destination.t(),
+          dictionary: Dictionary.t()
```
ManApart: Wait, so we no longer have steps? I was thinking that steps would include a step that includes a source (struct) etc. But are we actually removing those now? (Maybe steps are part of source / dest now, I need to keep reading).

I believe this means that define will need to create code / a component for each type of source struct, as opposed to a more generic 'step' component? May actually make the front end code cleaner, but also more verbose. Would be curious what Natty/Nick's thoughts are.

apps/definition_extract/lib/extract.ex
```diff
@@ -8,22 +8,20 @@ defmodule Extract do
...
         }
 
   defstruct version: 1,
```
ManApart: Should this now be version 2?

ManApart: Should this now be version 2?

apps/definition_extract/lib/extract/v1.ex
```diff
@@ -8,11 +8,10 @@ defmodule Extract.V1 do
       id: id(),
       dataset_id: required_string(),
       subset_id: required_string(),
-      destination: required_string(),
-      steps: spec(is_list() and not_nil?()),
+      source: impl_of(Source),
```
ManApart: Same question. Should this be version 2 now?

apps/definition/lib/definition/schema/type.ex
jeffgrunewald: checking if something is an implementation for a protocol, or what protocols have been implemented isn't anything new, but this would allow us to enforce Norm validation that the correct protocol has been implemented.

apps/definition_extract/lib/extract.ex
jeffgrunewald: since no one is using hindsight yet and our "prod" environment is technically our dev environment while our dev environment serves as our perf testing environment, I vote we kick this can and keep v1 as long as we don't have any deployments that have state we need to persist from one refactor to the next.

apps/decoder_gtfs/lib/decoder/gtfs.ex
```diff
@@ -0,0 +1,40 @@
...
+
+  defimpl Decoder do
+    def lines_or_bytes(_t), do: 2048
+
+    def decode(t, stream) do
+      stream
+      |> Enum.to_list()
+      |> List.flatten()
+      |> Enum.join()
+      |> TransitRealtime.FeedMessage.decode()
```
jmorris-pillar: Is this the message we always expect for gtfs? I  cant say I ever fully understood what we did with that since we only ever really had cota.

apps/broadway_configurator/lib/broadway_configurator.ex
ManApart: ^ This. :)

jeffgrunewald: What I mean is, after these deletions I don’t see where we’re starting broadway anymore which means we are either not doing gen_stage based consumption or we’re planning to roll our own.

apps/definition_kafka/lib/kafka/topic/source.ex
ManApart: Just reading through this I'm not sure what t is. Is `t` a `topic`?

apps/definition_load_broadcast/lib/load/broadcast.ex
```diff
@@ -34,7 +34,7 @@ defmodule Load.Broadcast.V1 do
       id: id(),
       dataset_id: required_string(),
       subset_id: required_string(),
-      source: required_string(),
+      source: impl_of(Source),
```
ManApart: 👍

apps/definition_load_persist/test/load/persist_test.exs
```diff
@@ -29,17 +29,13 @@ defmodule Load.PersistTest do
...
-          schema: [
-            %Dictionary.Type.String{name: "name"},
-            %Dictionary.Type.Integer{name: "age"}
-          ]
+          source: %Source.Fake{},
+          destination: "table"
         )
 
-      json = Jason.encode!(persist)
+      {:ok, serialized} = Brook.Serializer.serialize(persist)
```
ManApart: Reading the name of this test, I'd expect it to be testing something other than brook serialization of the persist struct. I'm not sure if the name is just misleading or if we should be using the store? It seems like json decoding/encoding or even using the Brook serialization is more testing Brook than it is testing something about persist.

apps/dlq/README.md
```diff
@@ -1,21 +1,21 @@
-# WriterDlq
+# Dlq
 
 **TODO: Add description**
```
ManApart: Please add some description!

apps/dlq/lib/dlq/server.ex
```diff
@@ -0,0 +1,57 @@
+defmodule Dlq.Server do
```
ManApart: Are these DLQ changes just a rename, or does it go deeper than that?

apps/protocol_decoder/README.md
```diff
@@ -0,0 +1,21 @@
+# ProtocolDecoder
+
+**TODO: Add description**
```
ManApart: Please add a description!

apps/protocol_destination/README.md
```diff
@@ -0,0 +1,21 @@
+# ProtocolDestination
+
+**TODO: Add description**
```
ManApart: Please add a description!

apps/protocol_extract_step/README.md
```diff
@@ -0,0 +1,21 @@
+# ProtocolExtractStep
+
+**TODO: Add description**
```
ManApart: Please add a description!

apps/protocol_source/README.md
```diff
@@ -1,21 +1,21 @@
-# TransformStep
+# ProtocolSource
 
 **TODO: Add description**
```
ManApart: Description please!

apps/definition_kafka/lib/kafka/topic/source.ex
bbalser: yeah, I am going to rename those variables

apps/decoder_gtfs/lib/decoder/gtfs.ex
bbalser: yes

apps/definition_deadletter/.gitignore
bbalser: git is just confused

apps/definition_extract/lib/extract.ex
bbalser: So my current thought was the extraction pipeline( the steps) becomes one implementation of a source.  So we are not getting rid of steps but moving them to a source implementation.  However the idea of reversing it and making a source of of the steps is really interesting.  I would love to brainstorm around this today or Monday.

ManApart: I'd like that!

apps/definition_kafka/lib/kafka/topic/destination.ex
```diff
@@ -0,0 +1,174 @@
...
+  defp store_connection(topic, connection) do
+    table = table_name(topic)
+
+    case :ets.whereis(table) do
+      :undefined -> :ets.new(table, [:named_table, :protected])
+      _ -> :ok
+    end
+
+    :ets.insert(table, {self(), connection})
+  end
```
jeffgrunewald: since this is a protected table, no one else can write to it but the process that was just init'ed; since this is in a handle_continue where would this table have come from if it were to already exist?

apps/definition_load_persist/test/load/persist_test.exs
jeffgrunewald: since brook serializer is a protocol, this validates that each custom struct maintains that contract across versions.

apps/decoder_gtfs/lib/decoder/gtfs.ex
bbalser: I am not a gtfs expert, but I believe that this is always the message.

apps/dlq/lib/dlq/server.ex
bbalser: It goes deeper, while brainstorming around source/destination, we wanted to move ownership of Json encoding/decoding the data into the source and destination implementations.  That meant that these implementations needed to DLQ.  We may refactor this again after learning more as we implement source and destinations.

apps/definition_kafka/lib/kafka/topic/destination.ex
jdenen: That... is a fair point. I copied this logic from some testing modules without really thinking it through. There, it makes sense. This does not.

apps/testing/lib/decoder/noop.ex
```diff
@@ -0,0 +1,16 @@
+defmodule Decoder.Noop do
```
jdenen:
```suggestion
defmodule Decoder.Fake do
```
Why? Why didn't you name it consistently?!

bbalser: lets go with, I did it on purpose to drive you nuts.  But in reality I just didn't think of it.

apps/decoder_json/lib/decoder/json.ex
```diff
@@ -0,0 +1,31 @@
+defmodule Decoder.Json do
+  use Definition, schema: Decoder.Json.V1
+
+  @derive Jason.Encoder
+  defstruct chunk_size: 1000
+
+  defimpl Decoder do
+    def lines_or_bytes(_t), do: :line
```
ManApart: Why do both json and json_lines return `:line` ?

apps/definition_dictionary/test/dictionary/type/latitude_test.exs
```diff
@@ -2,22 +2,6 @@ defmodule Dictionary.Type.LatitudeTest do
   use ExUnit.Case
   import Checkov
 
-  describe "new/1" do
```
ManApart: Why are all these `type.new/1` tests deleted? Are we doing/testing this elsewhere now?

apps/definition_kafka/lib/kafka/topic.ex
```diff
@@ -0,0 +1,64 @@
...
+
+  defp fix_partitioner(partitioner), do: partitioner
+
+  defp fix_endpoint([host, port]) when is_binary(host) and is_integer(port) do
+    {String.to_atom(host), port}
+  end
+
+  defp fix_endpoint(endpoint), do: endpoint
+
+  defimpl Source do
```
ManApart: Why implement and then delegate here instead of just implementing at the source/dest module? I can see it making sense to do this as a facade / high level interface, but I could also see just grabbing the source and or destination module if I needed it.

apps/protocol_destination/lib/destination.ex
```diff
@@ -0,0 +1,13 @@
...
+  def start_link(t, context)
+
+  @spec write(t, messages :: list(term)) :: :ok | {:error, term}
+  def write(t, messages)
+
+  @spec stop(t) :: :ok
+  def stop(t)
+
+  @spec delete(t) :: :ok | {:error, term}
+  def delete(t)
```
ManApart: Glad to see we have both a stop _and_ a delete.

apps/decoder_json/lib/decoder/json.ex
bbalser: That function has a different meaning, we probably should rename it, the purpose of that function is to tell the extractors if the decoder expects lines of text of raw bytes.   Which reminds me I don't think I am calling this function yet.

apps/definition_kafka/lib/kafka/topic.ex
bbalser: Because we implementing Source for the Kafka.Topic struct, so it felt natural that he imp be there.  We could move it to the the source.ex file, but the impl is still for Kafka.Topic.  My vote is to leave it where it is.

apps/definition_dictionary/test/dictionary/type/latitude_test.exs
jdenen: Turns out these tests were never useful. They would never fail. They did a pattern match with variables, but never pinned those variables.

So we've gone quite a while with these useless tests and not once has it come up. The only reason I found it was because I expected one to fail and it didn't. 

Given their general uselessness, I deleted them all.

ManApart: Lol. Nice. Good work, Boss.

PR #205, opened 2020-04-01
Switch to use Postgres instead of Redis to store state

kennyatpillar:


apps/service_broadcast/lib/broadcast/init.ex
```diff
@@ -1,16 +1,22 @@
...
   def on_start(state) do
-    Broadcast.Stream.Store.get_all!()
-    |> Enum.reject(&is_nil/1)
-    |> Enum.reject(&Broadcast.Stream.Store.done?(&1))
-    |> Enum.each(fn load ->
-      Broadcast.Stream.Supervisor.start_child({Broadcast.Stream, load: load})
-    end)
-
-    {:ok, state}
+    retry with: constant_backoff(100) do
```
jeffgrunewald, line 8: have a look at AnnotatedRetry, it does the same thing with mostly the same arguments but in a cleaner way. It's an app within the umbrella.

jeffgrunewald: Also, looking at everything that pipe is doing, I'd just create a separate function to retry connecting to the database to "list tables" or similar and retrying only that one. Then make a single call to the `get_all` that pipes through three other things we probably don't want to retry.

apps/initializer/lib/initializer.ex
```diff
@@ -26,10 +26,16 @@ defmodule Initializer do
           Map.new(init_arg)
           |> Map.put(:supervisor_ref, supervisor_ref)
 
+        {:ok, state}
```
jeffgrunewald, line 29: handle_continue will never be called unless `init/1` returns `{:ok, state, {:continue, :something}}`

apps/initializer/lib/initializer.ex
```diff
@@ -26,10 +26,16 @@ defmodule Initializer do
...
+      end
+
+      def handle_continue(:init, state) do
         case on_start(state) do
           {:ok, new_state} -> {:ok, new_state}
           {:error, reason} -> {:stop, reason}
         end
+
+        {:no_reply, state}
```
jeffgrunewald, line 38: get rid of this and just have the case statement return the `{:noreply, new_state}` or `{:stop, reason}`. The stop should cause things upstream to react since we're not expecting the Init servers to stop.

apps/service_broadcast/lib/broadcast/init.ex
```diff
@@ -1,16 +1,22 @@
...
+      |> Enum.reject(&is_nil/1)
+      |> Enum.reject(&Broadcast.Stream.Store.done?(&1))
+      |> Enum.each(fn load ->
+        Broadcast.Stream.Supervisor.start_child({Broadcast.Stream, load: load})
+      end)
+    after
+      _ ->
+        {:ok, state}
+    else
+      _ -> {:stop, "Could not read state from store", state}
```
jeffgrunewald, line 19: this else should still return an `{:error, reason}` tuple and then the case statement in the `Init` server handle_continue will translate it into the correct genserver return value.

apps/service_broadcast/lib/broadcast/init.ex
kennyatpillar: Thanks Jeff! I didn't realize AnnotatedRetry was a separate thing. Also, does the Brook Postgres driver have a good way to do "list tables" or would that be a case of me running a SQL statement to list the tables? I don't like the 2nd option because it feels like the app may know too many implementation details of the underlying storage layer, but if that's the cleanest way, I can do that.

apps/initializer/lib/initializer.ex
kennyatpillar: OK. I misunderstood how that worked.

apps/service_broadcast/lib/broadcast/init.ex
jeffgrunewald: this could be tricky since postgres doesn't really have a way to "list tables" in SQL (the command is something like `/dt`) so that's a limitation of the underlying Postgrex driver. You'll probably need to run the `get_all` query on its own until it returns a success and then send that through the pipeline.

config/releases.exs
```diff
@@ -78,9 +78,25 @@ config :service_receive, Receive.Application,
...
+        postgrex_args: [
+          hostname: "localhost",
+          username: "receive_app_user",
+          password: "receive123",
+          database: "receive_app_state"
+        ],
+        event_limits: %{
+          # "data:something" => 100,
+          # "data:something:else" => 50
+        }
```
jeffgrunewald, line 98: i think this block can just be omitted until we want to enforce limits on any events.

config/releases.exs
```diff
@@ -106,8 +122,20 @@ config :service_gather, Gather.Application,
...
+        postgrex_args: [
+          hostname: "localhost",
+          username: "gather_app_user",
+          password: "gather123",
+          database: "gather_app_state"
+        ],
+        event_limits: %{
+          # "data:something" => 100,
+          # "data:something:else" => 50
+        }
```
jeffgrunewald, line 137: same here

config/releases.exs
```diff
@@ -140,8 +168,20 @@ config :service_broadcast, Broadcast.Application,
...
+        postgrex_args: [
+          hostname: "localhost",
+          username: "broadcast_app_user",
+          password: "broadcast123",
+          database: "broadcast_app_state"
+        ],
+        event_limits: %{
+          # "data:something" => 100,
+          # "data:something:else" => 50
+        }
```
jeffgrunewald, line 183: same here

config/releases.exs
```diff
@@ -179,8 +219,20 @@ config :service_persist, Persist.Application,
...
+        postgrex_args: [
+          hostname: "localhost",
+          username: "persist_app_user",
+          password: "persist123",
+          database: "persist_app_state"
+        ],
+        event_limits: %{
+          # "data:something" => 100,
+          # "data:something:else" => 50
+        }
```
jeffgrunewald, line 234: same here

config/releases.exs
```diff
@@ -205,8 +257,20 @@ config :service_orchestrate, Orchestrate.Application,
...
+        postgrex_args: [
+          hostname: "localhost",
+          username: "orchestrate_app_user",
+          password: "orchestrate123",
+          database: "orchestrate_app_state"
+        ],
+        event_limits: %{
+          # "data:something" => 100,
+          # "data:something:else" => 50
+        }
```
jeffgrunewald, line 272: same here

config/releases.exs
```diff
@@ -236,8 +300,20 @@ config :service_acquire, Acquire.Application,
...
+        postgrex_args: [
+          hostname: "localhost",
+          username: "acquire_app_user",
+          password: "acquire123",
+          database: "acquire_app_state"
+        ],
+        event_limits: %{
+          # "data:something" => 100,
+          # "data:something:else" => 50
+        }
```
jeffgrunewald, line 315: same here

config/releases.exs
```diff
@@ -273,8 +349,20 @@ config :service_define, Define.Application,
...
+        postgrex_args: [
+          hostname: "localhost",
+          username: "define_app_user",
+          password: "define123",
+          database: "define_app_state"
+        ],
+        event_limits: %{
+          # "data:something" => 100,
+          # "data:something:else" => 50
+        }
```
jeffgrunewald, line 364: same here

config/releases.exs
```diff
@@ -297,8 +385,20 @@ config :service_profile, Profile.Application,
...
+        postgrex_args: [
+          hostname: "localhost",
+          username: "profile_app_user",
+          password: "profile123",
+          database: "profile_app_state"
+        ],
+        event_limits: %{
+          # "data:something" => 100,
+          # "data:something:else" => 50
+        }
```
jeffgrunewald, line 400: same here

apps/initializer/lib/initializer.ex
kennyatpillar: Fixed

apps/service_broadcast/lib/broadcast/init.ex
kennyatpillar: Fixed for this case. I did the same type of set up for all the apps. I'll get this one working and then do the rest.

apps/service_broadcast/lib/broadcast/init.ex
```diff
@@ -1,16 +1,22 @@
...
+    case(store) do
+      {:error, _} -> {:error, "Broadcast failed to read state from store during startup"}
+      _ ->
+        store |> Enum.reject(&is_nil/1)
+        |> Enum.reject(&Broadcast.Stream.Store.done?(&1))
+        |> Enum.each(fn load ->
+          Broadcast.Stream.Supervisor.start_child(load)
+        end)
+        {:ok, state}
+    end
```
jeffgrunewald, line 20:
```suggestion
    safe_retrieve()
    |> Enum.reject(&is_nil/1)
    |> Enum.reject(&Broadcast.Stream.Store.done?(&1))
    |> Enum.each(fn load ->
           Broadcast.Stream.Supervisor.start_child(load)
         end)

    {:ok, state}
  end
    
  @retry with: constant_backoff(100) |> take(10)
  defp safe_retrieve()
    Broadcast.Stream.Store.get_all!()
  catch
    _, reason ->
      Logger.error(fn ->
        "State tables not ready; retrying..."
      end)
  end
```

apps/service_broadcast/lib/broadcast/init.ex
```diff
@@ -1,16 +1,22 @@
...
-    |> Enum.reject(&is_nil/1)
-    |> Enum.reject(&Broadcast.Stream.Store.done?(&1))
-    |> Enum.each(fn load ->
-      Broadcast.Stream.Supervisor.start_child(load)
-    end)
-
-    {:ok, state}
+    store = Broadcast.Stream.Store.get_all!()
+    case(store) do
+      {:error, _} -> {:error, "Broadcast failed to read state from store during startup"}
```
jeffgrunewald, line 12: the `{:error, reason}` tuple will never match here since `Broadcast.Stream.Store.get_all!()` is wrapping `Brook.get_all_values!()` which will only succeed and return the value or else raise an exception.

config/releases.exs
kennyatpillar: Removed.

config/releases.exs
kennyatpillar: That's what I get for copy/pasting an example.

apps/service_broadcast/lib/broadcast/init.ex
kennyatpillar: I'm just seeing this one. Brian suggested moving the Annotated Retry up to the main Initializer. I'll touch base and see if there's a way to combine the two approaches.

apps/service_broadcast/lib/broadcast/init.ex
kennyatpillar: I switched to Annotated Retry. Brian is suggesting putting that code in Initializer.ex so all the apps use it. Since every init right now is starting out with getting everything from the store, I think this is fine.

helm/Chart.yaml
```diff
@@ -8,10 +8,6 @@ home: https://github.com/inhindsight
 sources:
   - https://github.com/inhindsight/hindsight
 dependencies:
-  - name: redis
```
jeffgrunewald, line 11: do we want to leave this in and just set it to `enabled: false`?

kennyatpillar: The AC said rip out Redis, we aren't using it any more. Having two stores doesn't make sense.

helm/presets/local.yaml
```diff
@@ -28,9 +28,6 @@ strimzi:
       limits:
         cpu: 100m
         memory: 512Mi
-redis:
```
jeffgrunewald, line 31: same question: if we want to leave this in and just set it to `enabled: false` the  user can choose redis or postgres just by the configuration in their `releases.exs`

apps/initializer/lib/initializer.ex
```diff
@@ -8,6 +8,8 @@ defmodule Initializer do
     quote location: :keep do
       use GenServer
       use Retry
```
jeffgrunewald, line 10: i think you can remove the reference to `use Retry` if you've got `use Annotated.Retry`

apps/initializer/lib/initializer.ex
```diff
@@ -49,6 +55,18 @@ defmodule Initializer do
...
 
+      @retry with: constant_backoff(100) |> take(10)
+      defp do_on_start(state) do
+        case on_start(state) do
+          {:ok, new_state} ->
+            {:ok, new_state}
+
+          {:error, reason} ->
+            {:error,
+             "on_start for application init failed, retrying reading Brook state from store..."}
```
jeffgrunewald, line 66: i would make this error message more generic; might not necessarily be trying to initialize from Brook state.

jeffgrunewald: i would just change the error case to:
```
  error ->
    error
```
to let the underlying app's implementation of it's error message bubble up.

jeffgrunewald, line 66: If i recall @bbalser's suggestion, if you change `do_on_start/1` to a `with` block and _dont_ worry about an else block, the `{:error, reason}` coming out of the implementing application will automatically bubble up to the case statement in the initializer's handle_continue.

apps/service_broadcast/lib/broadcast/init.ex
kennyatpillar: Fixed for broadcast. I'll do the rest of the apps the same way if this looks good.

apps/initializer/lib/initializer.ex
kennyatpillar: This is still being used. Leaving it in for now.

apps/initializer/lib/initializer.ex
kennyatpillar: Fixed. Good call.

kennyatpillar: Fixed.

apps/initializer/lib/initializer.ex
kennyatpillar: Converted this to the new Annotated.Retry the other place it was being used here. Take a look at my updated version when you get a chance.

apps/platform_runner/scripts/local_postgres_setup/create_users_and_databases.sql
```diff
@@ -0,0 +1,25 @@
+--- This script is for setting up users and databases for local testing of the platform --
```
jeffgrunewald, line 1: can this move into a `test/support` directory to clearly indicate it's purpose?

apps/platform_runner/docker-compose.yml
```diff
@@ -27,6 +27,8 @@ services:
     logging:
       driver: none
     image: smartcitiesdata/postgres-testo:development
+    volumes:
+      - "./scripts/local_postgres_setup/:/docker-entrypoint-initdb.d"
```
jeffgrunewald, line 31: this will need to change if you go with the next suggestion

apps/service_broadcast/lib/broadcast/init.ex
```diff
@@ -4,13 +4,20 @@ defmodule Broadcast.Init do
...
+  defp restore_state_from_store(store) do
+    store
     |> Enum.reject(&is_nil/1)
     |> Enum.reject(&Broadcast.Stream.Store.done?(&1))
-    |> Enum.each(fn load ->
-      Broadcast.Stream.Supervisor.start_child(load)
-    end)
-
-    {:ok, state}
+    |> Enum.each(fn load -> Broadcast.Stream.Supervisor.start_child(load) end)
```
jeffgrunewald, line 21:
```suggestion
    |> Enum.each(&Broadcast.Stream.Supervisor.start_child/1)
```

apps/service_broadcast/lib/broadcast/stream/store.ex
```diff
@@ -35,9 +35,15 @@ defmodule Broadcast.Stream.Store do
...
+  @spec get_all() :: [Load.t()]
+  def get_all() do
+    case Brook.get_all_values(@instance, @collection) do
+      {:ok, results} ->
+        results |> Enum.map(&Map.get(&1, "load"))
+        {:ok, results}
+
+      {:error, _} ->
+        {:error, "Failed to get values from Broadcast store"}
+    end
```
jeffgrunewald, line 47: consider changing this to a `with` without the else block and pipe the result to the `Ok.map/2` function in the umbrella's Ok library so you can just pass the `{:ok, result}` straight to to Map like:

```suggestion
    with {:ok, _state} = results <- Brook.get_all_values(@instance, @collection) do
      Ok.map(results, &Map.get(&1, "load"))
    end
```
Plus, the way this is currently written, I don't think you're returning the thing you want since you pipe the results to the `Enum.map` call but then you don't re-bind results so you're just returning the results you go straight out of Brook pre-Map.

apps/service_gather/lib/gather/extraction/store.ex
```diff
@@ -39,9 +39,15 @@ defmodule Gather.Extraction.Store do
...
-    |> Enum.map(&Map.get(&1, "extract"))
+  @spec get_all() :: [Extract.t()]
+  def get_all() do
+    case Brook.get_all_values(@instance, @collection) do
+      {:ok, results} ->
+        results |> Enum.map(&Map.get(&1, "extract"))
+        {:ok, results}
+
+      {:error, _} ->
+        {:error, "Failed to get values from Gather store"}
```
jeffgrunewald, line 50: same suggestion as broadcast.

apps/service_persist/lib/persist/load/store.ex
```diff
@@ -53,9 +53,15 @@ defmodule Persist.Load.Store do
...
+  @spec get_all() :: [%Load{}]
+  def get_all() do
+    case Brook.get_all_values(@instance, @collection) do
+      {:ok, results} ->
+        results |> Enum.map(&Map.get(&1, "load"))
+        {:ok, results}
+
+      {:error, _} ->
+        {:error, "Failed to get values from Persist store"}
+    end
```
jeffgrunewald, line 65: same suggestions as broadcast store.

apps/service_receive/lib/receive/accept/store.ex
```diff
@@ -38,9 +38,15 @@ defmodule Receive.Accept.Store do
...
+  @spec get_all() :: [Accept.t()]
+  def get_all() do
+    case Brook.get_all_values(@instance, @collection) do
+      {:ok, results} ->
+        results |> Enum.map(&Map.get(&1, "accept"))
+        {:ok, results}
+
+      {:error, _} ->
+        {:error, "Failed to get values from Receive store"}
+    end
```
jeffgrunewald, line 50: same suggestion as broadcast store.

apps/platform_runner/scripts/local_postgres_setup/create_users_and_databases.sql
kennyatpillar: Can do!

apps/platform_runner/docker-compose.yml
kennyatpillar: Yep, already caught that. I thought my last commit included both path changes for local testing.

apps/service_broadcast/lib/broadcast/stream/store.ex
kennyatpillar: Do I need an else statement or are we good just letting this bubble up an error from the underlying call to Brook.get_all_values?

helm/templates/acquire/deployment.yaml
```diff
@@ -40,6 +40,14 @@ spec:
...
               protocol: TCP
           env:
+            - name: ACQUIRE_DB_HOST
+              value: {{ template "postgres.fullname" . }}
+            - name: ACQUIRE_DB_NAME
+              value: {{ .Values.presto.view_state_apps.acquire.db.name }}
+            - name: ACQUIRE_DB_USER
+              value: {{ .Values.presto.view_state_apps.acquire.db.user | required "app db.user must be set" }}
+            - name: ACQUIRE_DB_PASSWORD
+              value: {{ .Values.presto.view_state_apps.acquire.db.password | required "app db.password must be set" }}
```
jdenen, line 50: Are we referencing these anywhere in code?

kennyatpillar: Yes, in the releases.exs file to configure the Postgres driver connection.

jdenen:
I still prefer the `System.get_env/2` pattern, since it's both established throughout `releases.exs` and because I know it won't blow up `platform_runner`. If we switch to `System.fetch_env!/1`, running hindsight locally and running e2e tests become a PITA.

jdenen:
@kennyatpillar @jmorris-pillar @jeffgrunewald 

I'm working on some changes that dovetail nicely with this PR, and there's at least one sizeable change I'd like to make to this code. Since timing has become a concern, I'm going to take this PR off your hands @kennyatpillar and will get it merged to master alongside my changes.

jdenen:
Closing, as this PR is being pulled in by #243

PR #208, opened 2020-04-03
Ingress worldwide

jmorris-pillar:


helm/templates/letsencrypt-issuer.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: cert-manager.io/v1alpha2
+kind: ClusterIssuer
+metadata:
+  name: letsencrypt-issuer
+spec:
+  acme:
+    email: jdenen@pillartechnology.com
```
jdenen, line 7: Hmm. I don't _think_ this is the value I want here.

jdenen:
```suggestion
    email: johnson.denen@gmail.com
```
Unless there's a specific reason not to update, let's go with this.

Also, should this be configurable? (not a passive suggestion, I legit am unsure)

helm/templates/letsencrypt-issuer.yaml
```diff
@@ -0,0 +1,16 @@
...
+    server: https://acme-v02.api.letsencrypt.org/directory
+    privateKeySecretRef:
+      # Secret resource that will be used to store the account's private key.
+      name: hindsight-prod-tls-key
```
jdenen, line 11: Should `prod` be hardcoded into this name?

jeffgrunewald: I’d template it.

helm/templates/letsencrypt-issuer.yaml
jeffgrunewald: Should be able to template and require that.

helm/Chart.yaml
```diff
@@ -19,3 +19,6 @@ dependencies:
     version: 8.12.3
     repository: https://kubernetes-charts.storage.googleapis.com
     condition: prometheus-operator.enabled
+  - name: cert-manager
+    version: 0.14.1
+    repository: https://charts.jetstack.io
```
jeffgrunewald, line 24: if this is going to require a bunch of setup stuff in LetsEncrypt and obviously adding network overhead, would it be a good idea to make this another toggle (`cert-manager.enabled`) and then we could even default it to `true` but generally make it easy to disable TLS encryption for quick testing and local dev?

jeffgrunewald:
Can you also throw a comment in the README or maybe we need a `docs` directory to start collecting instructions and glossary and setup instructions to call out the need to manually install the cert-manager CRDs _before_ trying to deploy the helm chart with cert-manager enabled in case that upstream PR of theirs doesn't get merged soon or in case they only update it to support k8s 1.15+ and it takes us a while to upgrade our clusters?

helm/templates/letsencrypt-issuer.yaml
jmorris-pillar: I don't think its valuable to have the environment name in this key so to prevent unneeded complexity this will just be hindsight-tls-key in the respective environments.  It is just the private key for that environment.

helm/Chart.yaml
jmorris-pillar: Agreed

jdenen:
LGTM but I'll let @jeffgrunewald take another look too before I merge

helm/Chart.yaml
jeffgrunewald: 👍

PR #210, opened 2020-04-03
Documentation worldwide (1 of 3)

jdenen:
So maybe not so worldwide, but I'm trying to keep the number of files changed under 100 per PR.

### DIFF
--- a/apps/transformer/lib/transform/move_field.ex
+++ b/apps/transformer/lib/transform/move_field.ex
@@ -1,4 +1,15 @@
 defmodule Transform.MoveField do
+  @moduledoc """
+  `Transform.Step` impl to move or rename a field
+  in a dataset.
+
+  ## Configuration
+
+  * `from` - Required. Field to be moved. This can be a string or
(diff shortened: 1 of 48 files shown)


apps/transformer/lib/transform/move_field.ex
```diff
@@ -1,4 +1,15 @@
...
+  `Transform.Step` impl to move or rename a field
+  in a dataset.
+
+  ## Configuration
+
+  * `from` - Required. Field to be moved. This can be a string or
+  path. Ex: `"foo"` or `["foo", "bar", "baz"]`
+  * `to` - Required. Where to move the field. This can be a string
+  or path. Ex: `"abc"` or `["abc", "xyz"]`
+  """
```
jeffgrunewald, line 12: Would it make sense to rename this step to `Transform.MvField` to make the correlation to the unix `mv` command clearer without having to read the docstring? Just thinking if I'm looking for "rename" functionality and just glancing at the module names I might skip this one.

PR #214, opened 2020-04-07
View Data Definitions in Define

NathanielScottStevens:


jmorris-pillar:
Please squash and merge when merging this pr

Dockerfile
```diff
@@ -1,6 +1,13 @@
...
+COPY apps/service_define/assets/ /app
+WORKDIR /app/
+RUN npm install \
+  && npm run release
+
 FROM bitwalker/alpine-elixir-phoenix:1.9.4 as build
 COPY . /opt/app
-WORKDIR /opt/app
+WORKDIR /opt/app/
+COPY --from=npm_builder /app/dist/ apps/service_define/priv/static
```
jmorris-pillar, line 10: Why do we copy from dist to priv static?  What normally takes care of this in the release process that we cant just put this in a dist folder?

apps/service_define/assets/src/component/presentation/data-definition/DataDefinition.stories.tsx
```diff
@@ -0,0 +1,197 @@
...
+            struct_module_name: "Elixir.Extractor",
+            args: [
+                {
+                    key: "steps",
+                    type: ["list", PrimitiveArgumentType.module],
+                    value: [
+                        {
+                            struct_module_name: "Extract.Http.Get",
+                            args: [
+                                { key: "url", type: PrimitiveArgumentType.string, value: "http://www.source.com/the-data.csv"}
```
jmorris-pillar, line 21: Can we use example.com or something with .example as the tld just to be good denizens?

jmorris-pillar: The top-level domain example is explicitly intended to be used in documentation or other technical writing, when domain names are presented as examples in usage or presentation of concepts of the Domain Name System or the Internet.

apps/service_define/config/dev.exs
```diff
@@ -1,5 +1,20 @@
 use Mix.Config
 
+config :service_define, Define.Application,
+  kafka_endpoints: nil,
```
jmorris-pillar, line 4: Do you need this line?

apps/service_define/config/dev.exs
```diff
@@ -14,9 +29,7 @@ config :service_define, DefineWeb.Endpoint,
       "../priv/static/",
       cd: Path.expand("../assets", __DIR__)
     ]
-  ]
-
-config :service_define, DefineWeb.Endpoint,
+  ],
   live_reload: [
```
jmorris-pillar, line 33: Did you mean to delete this line that now puts live reload under the main app and not under the endpoint?

apps/service_define/dev_scripts/send_events.exs
```diff
@@ -0,0 +1,20 @@
...
+    dataset_id: id,
+    subset_id: "default",
+    destination: "success",
+    dictionary: [
+        Dictionary.Type.String.new!(name: "letter")
+    ],
+    steps: [
+        Extract.Http.Get.new!(
+        url: "http://localhost/file.csv",
+        headers: %{"content-length" => "5"}
```
jmorris-pillar, line 15: Why do we need to tell the content length on the request as a header?

apps/service_define/mix.exs
```diff
@@ -29,13 +30,21 @@ defmodule Define.MixProject do
...
+      {:extractor, in_umbrella: true},
       {:gettext, "~> 0.11"},
       {:jason, "~> 1.0"},
       {:metrics_reporter, in_umbrella: true},
       {:phoenix, "~> 1.4.15"},
       {:phoenix_pubsub, "~> 1.1"},
       {:plug_cowboy, "~> 2.0"},
       {:properties, in_umbrella: true},
+      # This should eventually be transformer once that has been created (like extractor)
+      {:transform_wkt, in_umbrella: true}
```
jmorris-pillar, line 47: comma

helm/presets/local.yaml
```diff
@@ -1,3 +1,6 @@
+define:
+  service:
+    type: NodePort
```
jmorris-pillar, line 3: Should this be a different service type than acquire?

helm/templates/define/deployment.yaml
```diff
@@ -0,0 +1,62 @@
...
+              protocol: TCP
+          env:
+            - name: KAFKA_ENDPOINTS
+              value: {{ template "hindsight.fullname" . }}-kafka-bootstrap:9092
+            - name: SECRET_KEY_BASE
+              valueFrom:
+                secretKeyRef:
+                  name: {{ template "hindsight.fullname" . }}-define-key-base
+                  key: secret_key_base
+            - name: REDIS_HOST
```
jmorris-pillar, line 47: One of you // kenny will need to make sure this gets reconciled to postgres

jmorris-pillar: i.e. the race is on.

helm/presets/local.yaml
```diff
@@ -1,3 +1,6 @@
+define:
```
jmorris-pillar, line 1: Are you also going ot need to have changes to aws.yaml?  We probably need this configured behind the ingress controller which I can help you with.  I'm okay if we do that in seperate pull requests to keep this one from getting longer.

.dockerignore
```diff
@@ -4,4 +4,7 @@ deps/
 .elixir_ls/
 .git/
 .github/
-helm/
\ No newline at end of file
+helm/
+apps/service_define/assets/node_modules/
+apps/service_define/assets/.cache/
+apps/service_define/assets/build/
```
jeffgrunewald, line 10: i take it there _are_ assets we want to upload so you can't just omit the entire `apps/service_define/assets/` dir?

apps/definition/lib/definition/schema/type.ex
```diff
@@ -44,7 +44,29 @@ defmodule Definition.Schema.Type do
...
+  end
+
+  @spec atom() :: spec
+  def atom() do
+    spec(is_atom())
+  end
+
+  @spec boolean() :: spec
+  def boolean() do
+    spec(is_boolean())
```
jeffgrunewald, line 67: i'll leave it to @bbalser or @jdenen since they've been deeper in Norm for longer: is it preferred to just directly call `spec(predicate_func())` directly when you're invoking Norm, or is it preferable to wrap these `spec/1` calls?

apps/definition_dictionary/lib/dictionary/type/string.ex
```diff
@@ -2,6 +2,12 @@ defmodule Dictionary.Type.String do
   use Definition, schema: Dictionary.Type.String.V1
   use Dictionary.JsonEncoder
 
+  @type t :: %__MODULE__{
+    version: integer,
+    name: String.t(),
+    description: String.t()
+  }
```
jeffgrunewald, line 9: many type definitions lost their lives to bring us that big source/destination refactor... 😞

apps/service_define/config/dev.exs
jeffgrunewald: we shouldn't; it's not explicitly defined in `config.exs` so it should default to `nil`

apps/service_define/config/dev.exs
```diff
@@ -1,5 +1,20 @@
...
+    driver: [
+      module: Brook.Driver.Default,
+      init_arg: []
+    ],
+    handlers: [Define.Event.Handler],
+    storage: [
+      module: Brook.Storage.Ets,
+      init_arg: []
+    ],
+    dispatcher: Brook.Dispatcher.Noop
```
jeffgrunewald, line 15: Since we're supplying these values "at runtime" in the tests by doing overrides using the `Properties` macro, are we including these here for running the application locally while doing development of the UI? if so, you should be able to remove the same settings in `test.exs` and supply the values in your specific tests by the `temp_env` module in the `:testing` umbrella app.

apps/service_define/lib/define/event/store.ex
```diff
@@ -0,0 +1,85 @@
...
+  def update_definition(%Load{} = data) do
+    update_definition_field(data, :load, &to_load_view/1)
+  end
+
+  def update_definition(%Extract{} = data) do
+    update_definition_field(data, :extract, &to_extract_view/1)
+  end
+
+  def update_definition(data) do
+    Logger.error("Got unexpected data definition update: #{inspect(data)}")
```
jeffgrunewald, line 23:
```suggestion
    Logger.error(fn -> "Got unexpected data definition update: #{inspect(data)}" end)
```

apps/service_define/lib/define/event/store.ex
```diff
@@ -0,0 +1,85 @@
...
+  end
+
+  defp update_definition_field(data, key, to_view_converter) do
+    get_or_create(data)
+    |> Map.put(key, to_view_converter.(data))
+    |> persist()
+  rescue
+    err ->
+      Logger.error("Unable to process transform event: #{inspect(err)}")
```
jeffgrunewald, line 32:
```suggestion
      Logger.error(fn -> "Unable to process transform event: #{inspect(err)}" end)
```

helm/presets/local.yaml
jeffgrunewald: this should be a `ClusterIP` type and include a path in the ingress manifest to reach Define directly from the web on 80.

.dockerignore
nbeach: This is correct. assets/ contains all the UI code, and these a just a few folders that get generated that we don't want.

Dockerfile
nbeach: Not sure what you mean by "just put this in a dist folder"? priv/static/ is where phoenix scoops up the static assets when released, so we first have to use a node docker image to build the UI, then copy those built files to the proper folder in the elixir image before the release is run. Is there a simplification we're missing here?

apps/service_define/dev_scripts/send_events.exs
NathanielScottStevens: We don't, it was just what we thought of first. Would you prefer we use something else?

apps/definition/lib/definition/schema/type.ex
NathanielScottStevens: Yeah, we were curious about this too. We were trying to follow precedent but may have understood that precedent incorrectly.

apps/service_define/config/dev.exs
NathanielScottStevens: I don't really know anything about the testing umbrella app. Why would we put our configuration there instead of with the apps?

apps/service_define/assets/src/component/presentation/data-definition/DataDefinition.stories.tsx
NathanielScottStevens: I almost suggested this while pairing and then thought "no, that's a little too pedantic". I'll put it in for you though Jessie.

apps/service_define/config/dev.exs
NathanielScottStevens: `live_reload` is still under the endpoint.  This diff is just really confusing I think.

helm/presets/local.yaml
NathanielScottStevens: I will add that now. Although I'm not sure what needs to be done with the ingress controller. I'll need your help there for sure.

jeffgrunewald: Nothing needs done to the ingress _controller_ but you need to add another path to `Values.ingress.paths` in the root values.yaml file and change the define service to a ClusterIP (not required but more restricted than NodePort and obviates the need for NodePort)

apps/service_define/config/dev.exs
jeffgrunewald: Testing gives you a module called `Temp.Env` that lets you set app env configs in a specific test to avoid blanket configurations across the test environment and keeps the `test.exs` and `integration.exs` files much cleaner. It takes the app env stuff that we're setting via the Properties library (also umbrella) and configuring them at runtime, both defaulting to prod settings and overriding only as needed in the tests that require it.

Look at the service_persist's `lib/event/handler.ex` and its associated test to see how we're defining configuration and then overriding it as needed in a test.

apps/service_define/mix.exs
```diff
@@ -11,6 +11,7 @@ defmodule Define.MixProject do
       lockfile: "../../mix.lock",
       elixir: "~> 1.5",
       elixirc_paths: elixirc_paths(Mix.env()),
+      elixirc_options: [debug_info: true],
```
jeffgrunewald, line 14: is this something we want turned on permanently?

apps/service_define/config/dev.exs
NathanielScottStevens: It seems simpler to just to provide them through the config for now until we have a reason to override them in a test.

apps/service_define/lib/define/event/store.ex
```diff
@@ -0,0 +1,85 @@
...
+  end
+
+  defp update_definition_field(data, key, to_view_converter) do
+    get_or_create(data)
+    |> Map.put(key, to_view_converter.(data))
+    |> persist()
+  rescue
+    err ->
+      Logger.error(fn -> "Unable to process transform event: #{inspect(err)}" end)
+      discard()
```
jdenen, line 33:
```suggestion
      :discard
```
I think we've taken the abstraction too far.

apps/service_define/lib/define/event/handler.ex
```diff
@@ -0,0 +1,18 @@
...
+    Define.Event.Store.update_definition(extract)
+  end
+
+  def handle_event(%Brook.Event{type: transform_define(), data: %Transform{} = transform}) do
+    Define.Event.Store.update_definition(transform)
+  end
+
+  def handle_event(%Brook.Event{type: load_start(), data: %Load{} = load}) do
+    Define.Event.Store.update_definition(load)
+  end
```
jdenen, line 17:
```suggestion
  @events [extract_start(), transform_define(), load_start()]
  
  def handle_event(%Brook.Event{type: event, data: data}) when event in @events do
    Define.Event.Store.update_definition(data)
  end
```
Just seems like a place to DRY up.

apps/service_define/lib/define/event/store.ex
```diff
@@ -0,0 +1,85 @@
...
+    update_definition_field(data, :load, &to_load_view/1)
+  end
+
+  def update_definition(%Extract{} = data) do
+    update_definition_field(data, :extract, &to_extract_view/1)
+  end
+
+  def update_definition(data) do
+    Logger.error(fn -> "Got unexpected data definition update: #{inspect(data)}" end)
+  end
```
jdenen, line 24: What's the spec of this function? Why is an unexpected input returning the same result as a good input?

apps/service_define/lib/define/event/store.ex
```diff
@@ -0,0 +1,85 @@
...
+
+  defp update_definition_field(data, key, to_view_converter) do
+    get_or_create(data)
+    |> Map.put(key, to_view_converter.(data))
+    |> persist()
+  rescue
```
jdenen, line 30: Not a big fan of using rescue like this. Can we refactor to use `:ok/:error` tuples throughout?

apps/service_define/lib/define/event/store.ex
jdenen: See my question about `:ok/:error` below. Same deal feels like it should apply here.

.dockerignore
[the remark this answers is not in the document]
jdenen: I honestly don't know, so feel free to ignore.

apps/service_define/lib/define/typespec_analysis.ex
```diff
@@ -0,0 +1,68 @@
+defmodule Define.TypespecAnalysis do
```
jdenen, line 1: Definitely document this.

apps/service_define/lib/define/typespec_analysis.ex
```diff
@@ -0,0 +1,68 @@
...
+  import Logger
+
+  def find() do
```
jdenen, line 4:
```suggestion
  def find do
```
Also add a spec, please.

apps/service_define/lib/define/typespec_analysis.ex
```diff
@@ -0,0 +1,68 @@
...
+
+      :error ->
+        Logger.error("Unable to get types for #{module}")
+        raise(ArgumentError, "Unable to get types for #{module}")
+    end
+  end
+
+  defp extract_fields([]) do
+    []
+  end
```
jdenen, line 33:
```suggestion
  defp extract_fields([]), do: []
```

jdenen: I'm getting real nit-picky, I know. Sorry.

apps/service_define/lib/define/typespec_analysis.ex
```diff
@@ -0,0 +1,68 @@
...
+  defp extract_field_type({:type, _, _, [{_, _, name}, {_, _, type}]}), do: {name, type, nil}
+
+  defp extract_field_type({:type, _, _, [{_, _, name}, {_, _, type, sub_typespec}]}),
+    do: {name, type, sub_typespec}
+
+  defp to_simple_type({name, :list, [{_, _, [{_, _, String}, _, _]}]}),
+    do: {to_string(name), {"list", "string"}}
+
+  defp to_simple_type({name, :list, [{_, _, type, _}]}),
+    do: {to_string(name), {"list", to_string(type)}}
```
jdenen, line 52: Oof, these are hard to parse. Can we make them proper `do/end` style functions. That'll at least break what parts I'm supposed to care about from the ones I'm not.

apps/service_define/lib/define/typespec_analysis.ex
```diff
@@ -0,0 +1,68 @@
...
+  defp to_simple_type({name, typespec, _}) do
+    type =
+      case typespec do
+        [{_, _, String}, _, _] -> "string"
+        [{_, _, Dictionary}, _, _] -> "dictionary"
+        _ -> typespec
+      end
+
+    {to_string(name), to_string(type)}
+  end
```
jdenen, line 63: I think this could be simplified by pulling parsing logic out of the function head and into a private function:

```elixir
defp to_simple_type({name, typespec, details}) do
  type = 
    case typespec do
      :list -> {"list", list_item_simple_type(details)}
      [{_, _, String} | _] -> "string"
      [{_, _, Dictionary} | _] -> "dictionary"
      _ -> to_string(typespec)
    end

  {to_string(name), type}
end

defp list_item_simple_type([{_, _, [{_, _, String} | _]}]), do: "string"
defp list_item_simple_type([{_, _, type, _}]), do: to_string(type)
```

jdenen: Hmm. Even after writing that out I believe this should be simpler. Maybe we should parse the `{name, typespec, details}` tuple into a struct and have helper functions that return field and simple types.

It feels complicated to parse field and simple types from nested and convoluted tuples/lists.

apps/service_define/lib/define/view_state.ex
```diff
@@ -20,9 +23,8 @@ defmodule Define.ViewState do
   end
 
   @impl true
-  def handle_call({:event, type, payload}, _from, state) do
-    new_state = update_state(state, type, payload)
-    {:reply, new_state, new_state}
+  def handle_call({:event, _type, _payload}, _from, state) do
+    {:reply, state, state}
```
jdenen, line 27: I assume we continue to keep this module for what we know to be coming? We might want to document that for now.

apps/service_define/test/define/typespec_analysis_test.exs
```diff
@@ -0,0 +1,23 @@
...
+      "map_type" => "map",
+      "dictionary_type" => "dictionary"
+    }
+
+    assert expected == TypespecAnalysis.get_types(TypespecAnalysisTestModule)
+  end
+
+  test "returns empty map when a typespec is not defined" do
+    assert %{} == TypespecAnalysis.get_types(TypespecAnalysisTestModuleNoSpec)
+  end
```
jdenen, line 22:
```suggestion
  describe "get_types/1" do
    test "returns all types from the typespec" do
      expected = %{
        "integer_type" => "integer",
        "string_type" => "string",
        "atom_type" => "atom",
        "float_type" => "float",
        "string_list_type" => {"list", "string"},
        "integer_list_type" => {"list", "integer"},
        "map_type" => "map",
        "dictionary_type" => "dictionary"
      }

      assert expected == TypespecAnalysis.get_types(TypespecAnalysisTestModule)
    end

    test "returns empty map when a typespec is not defined" do
      assert %{} == TypespecAnalysis.get_types(TypespecAnalysisTestModuleNoSpec)
    end
  end
```

This is more or less the pattern we're following for grouping tests around a single function.

Dockerfile
jmorris-pillar: Probably just somethign I'm missing locally.  I figure if you ran it locally you run something like npm run release inside the assets folder and dont have to then copy them to priv/static, but im guessing theres some nuance with distillery we're chasing?

apps/service_define/assets/package.json
```diff
@@ -8,7 +8,7 @@
     "compile": "tsc --noEmit",
     "ci": "npm run compile && npm run test:once && npm run lint",
     "storybook": "start-storybook -s ./public",
-    "deploy": "parcel build src/index.html --out-dir ../priv/static/",
+    "release": "parcel build src/index.html --out-dir ./dist",
```
jmorris-pillar, line 11: It is possible this is the missing piece to my dockerfile question.

apps/service_define/config/dev.exs
jmorris-pillar: sure is, thanks lol

apps/service_define/lib/define/event/handler.ex
jmorris-pillar: Do you lose a secondary check that those event types have the right structs though?  Does that matter?

jdenen: Take a look at `apps/definition_events/lib/events.ex`.

This macro creates send_* functions for the events that ensure event type and event struct are locked together.

apps/service_define/config/dev.exs
NathanielScottStevens: So that we can run the app locally without any dependencies. I guess we could add it to the top level umbrella base config but that seems weird to me.

apps/definition/lib/definition/schema/type.ex
```diff
@@ -44,7 +44,29 @@ defmodule Definition.Schema.Type do
     end)
   end
 
-  @spec access_path() :: spec
+  @spec map() :: spec
+  def map() do
+    spec(fn
+      %{__struct__: _} -> false
+      x -> is_map(x)
```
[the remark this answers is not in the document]
NathanielScottStevens: Yes, we handle structs and maps differently when we generate the UI

apps/service_define/lib/define/event/store.ex
NathanielScottStevens: Yeah, I don't care for this either.

apps/service_define/lib/define/event/store.ex
NathanielScottStevens: I originally took us down that path but then Nick pointed out that we were adding a lot of extra work/verbosity by doing so which I found myself agree with. I forget all the particulars about why it got so gross at this point though.

apps/service_define/mix.exs
NathanielScottStevens: No, good catch

apps/service_define/lib/define/event/store.ex
NathanielScottStevens: I'm not quite sure I'm following what you're thinking here. By an unexpected input do you mean an unexpected event? `update_definition_field` ends up storing the event in the view state whereas the catch all does not. Are you wanting us to bring the storing in the viewstate up to these top level functions to make that more apparent?

jdenen: `Logger.error/1` returns `:ok` was what I was getting at.

If a valid definition is passed to `update_definition` but fails to persist, we log an error and return `:discard`.

If an invalid definition is passed to `update_definition`, we log an error and return `:ok`.

apps/service_define/lib/define/event/store.ex
jdenen: So we're going to use exceptions for control flow because of verbosity?

helm/presets/aws.yaml
```diff
@@ -101,6 +104,14 @@ presto:
 cert_manager:
   enabled: true
 
+global:
```
jmorris-pillar, line 107: Do you need to add this.  I think this reverts a change jeff recently made?

helm/presets/local.yaml
```diff
@@ -66,7 +69,16 @@ presto:
       annotations:
         nginx.ingress.kubernetes.io/use-regex: "true"
       hosts: [storage.hindsight.local/minio.*]
-
+global:
```
jmorris-pillar, line 72: why do we need to add this?

helm/presets/aws.yaml
NathanielScottStevens: I'm guessing we just messed up the merge

PR #216, opened 2020-04-10
Documentation worldwide (part 2)

jdenen:
With this and #210, all things touched by #188, #206, and #209 should be documented.


jdenen:
One more PR to come, which should close #49 and #50. With that PR will also come `credo` to make sure I never have to do all this documentation at one time, ever again.

PR #220, opened 2020-04-14
adding vernemq chart, regrouping values

jeffgrunewald:
Adds VerneMQ cluster
- defaults to single node in local, 3 nodes in aws
- regroups the subchart values to make them easier to sort out from hindsight variables
- cleans up the install script/dependencies
- NOTE leaves Verne turned off in aws for the time being until it's needed/development of consumer progresses further

### DIFF
--- a/helm/presets/aws.yaml
+++ b/helm/presets/aws.yaml
@@ -25,7 +25,10 @@ global:
     minio:
       enabled: false
 
-# Strimzi subchart
+# Subchart configuration is namespaced under the subchart's name
+# as a top-level key. For example, the Certificate Manager's subchart is
+# configured under the cert_manager key.
+
 strimzi:
   kafka:
     storage:
@@ -53,12 +56,10 @@ strimzi:
         cpu: 250m
         memory: 512Mi
 
-# Redis subchart
 redis:
   enabled: false
   externalAddress: ""
 
-# Nginx subchart
 nginx-ingress:
   controller:
     service:
@@ -66,7 +67,6 @@ nginx-ingress:
         service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout: "3600"
       type: LoadBalancer
 
-# Prometheus subchart
 prometheus-operator:
   enabled: true
   prometheus:
@@ -89,7 +89,6 @@ prometheus-operator:
     ingress:
       enabled: true
 
-# Presto subchart
 presto:
   postgres:
     enable: false
@@ -99,11 +98,9 @@ presto:
       user: ""
       password: ""
 
-# Cert Manager subchart
 cert_manager:
   enabled: true
 
-# VerneMQ subchart
 vernemq:
   enabled: false
   persistentVolume.size: 25Gi
--- a/helm/presets/aws.yaml
+++ b/helm/presets/aws.yaml
@@ -16,6 +16,16 @@ resources:
   limits:
     cpu: 500m
     memory: 512Mi
+
+global:
+  objectStore:
+    region: "us-east-2"
+    accessKey: ""
+    accessSecret: ""
+    minio:
+      enabled: false
+
+# Strimzi subchart
 strimzi:
   kafka:
     storage:
@@ -42,15 +52,21 @@ strimzi:
       limits:
         cpu: 250m
         memory: 512Mi
+
+# Redis subchart
 redis:
   enabled: false
   externalAddress: ""
+
+# Nginx subchart
 nginx-ingress:
   controller:
     service:
       annotations:
         service.beta.kubernetes.io/aws-load-balancer-connection-idle-timeout: "3600"
       type: LoadBalancer
+
+# Prometheus subchart
 prometheus-operator:
   enabled: true
   prometheus:
@@ -72,6 +88,8 @@ prometheus-operator:
   grafana:
     ingress:
       enabled: true
+
+# Presto subchart
 presto:
   postgres:
     enable: false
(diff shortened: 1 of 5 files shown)


helm/presets/aws.yaml
```diff
@@ -42,15 +52,21 @@ strimzi:
       limits:
         cpu: 250m
         memory: 512Mi
+
+# Redis subchart
```
jdenen, line 56: These type of comments seem like noise.

helm/presets/aws.yaml
```diff
@@ -81,13 +99,16 @@ presto:
...
 
+# Cert Manager subchart
 cert_manager:
   enabled: true
 
-global:
-  objectStore:
-    region: "us-east-2"
-    accessKey: ""
-    accessSecret: ""
```
jdenen, line 91: If we're removing all references to these, then they should be validated via the `required` macro. That way it can't be a surprise.

scripts/install
```diff
@@ -28,8 +28,7 @@ function check_strimzi {
 
 function deps {
     info "Installing Helm dependencies"
-    helm repo add google https://kubernetes-charts.storage.googleapis.com
```
jdenen, line 31: Anyone installing hindsight for the first time will need this line for the script to work, won't they?

helm/presets/aws.yaml
jdenen: Also, you're breaking an implicit contract since we wouldn't be defaulting to us-east-2. Maybe that's fine, especially this early, but it's worth noting.

The documentation (README, etc) probably needs updated to reflect these changes.

scripts/install
```diff
@@ -40,7 +39,7 @@ function strimzi {
         info "Installing Strimzi"
         helm upgrade strimzi-kafka-operator strimzi/strimzi-kafka-operator \
              --install \
-             --version 0.16.2 \
+             --version 0.17.0 \
```
jdenen, line 42: Is this a safe upgrade? Strimzi breaks contracts pretty regularly.

helm/presets/aws.yaml
jeffgrunewald: I wanted to call out how values for a subchart are overridden from a parent chart since that is not obvious to everyone.

jeffgrunewald: _someone_ who first stood up the hindsight chart didn't have the presto chart properly referenced viz-a-viz overriding subchart values from a top-level `values.yaml` file and we were just getting lucky based on that chart's defaults. 😁

helm/presets/aws.yaml
jeffgrunewald: This is still in the file; just moved up above the subchart values references.

scripts/install
jeffgrunewald: /shrug
it's not critical

scripts/install
jeffgrunewald: changing the `helm dependency build` to `helm dependency update` takes care of this. if we were installing anything from the stable repo via `helm upgrade --install foo bar.com` the way we do with Strimzi we would, but since the only things we need from the stable repo are being installed as _part of hindsight_ dependencies this isn't required.

jeffgrunewald: to put it a different way, if we're _only_ including charts from that repo as explicit dependencies of hindsight's chart, this separate line isn't necessary

scripts/install
jeffgrunewald: do we have any data we care about if we're failing forward here though?

scripts/install
jdenen: 👍

jdenen: Thanks for the clarification

jeffgrunewald:
The team seemed to want to have a broker cluster to test against for developing the mqtt interface in gather.
That being said, that's more or less why I only have Verne enabled in the local presets at the moment; to make developing and testing easier. The simulator app can already interact with mqtt so we have sample source data to feed into the cluster by the end of the this week at the latest and the next cards on the board that have been groomed and are ready to play are to implement mqtt in gather

scripts/install
jdenen: You'll answer to @jmorris-pillar when you break prod, is all I'm saying.

helm/presets/aws.yaml
jdenen: I get fixing the organization/structure of the variables.

If you want to have some commentary on the reasoning of the organization/structure, add a comment to the top. Something like:

```
# Subchart configuration is namespaced under the subchart's name
# as a top-level key. For example, the Certificate Manager's subchart is 
# configured under the cert_manager key.
```

To me, that comment is valuable and the many comments of `$keyname subchart` are not.

helm/presets/aws.yaml
jdenen: Ah, damn diff got me. Thanks.

jdenen:
Gotcha. Makes sense.

helm/presets/aws.yaml
jeffgrunewald: fair enough; i can get behind that.

jeffgrunewald: Just pushed a fix; let me know what you think.

helm/presets/local.yaml
```diff
@@ -49,13 +66,13 @@ presto:
...
-    accessSecret: hindsightAccessSecret
-    minio:
-      enabled: true
-      port: 9000
-      scheme: "http://"
+
+vernemq:
+  enabled: true
+  service:
+    type: LoadBalancer
```
jmorris-pillar, line 73: Why does verne need to be a load balancer but. acquire does not?  I think you've explained this to me once before but I forgot lol

jeffgrunewald: Since Acquire is over HTTP we can keep that an internal-only ClusterIP and have an ingress object route the traffic from the Nginx LoadBalancer service to the Acquire ClusterIP service. Since MQTT and MQTTS are not over HTTP, they'd need to be LoadBalancer type-services to be reachable outside of the cluster

PR #222, opened 2020-04-17
Refactors to view state management

jdenen:
This is the first PR of many to change how services manage view state.

Currently, we rewrite a LOT of the same boilerplate across services to persist/update/delete view states. And we put all that code in different files. And there's small inconsistencies between them.

### Management.ViewState

This PR adds `Management.ViewState` which mixes in to modules that encapsulate the view state of a single collection. So, one event object type == one `Management.ViewState`-using collection module.

### Gather

This PR includes changes to Gather to use the new view state stuff. This also changes how it handles `extract_end` and `dataset_delete` events to work with the new functions.

This PR only includes changes to Gather, since each service can be updated independently.

### Postgres

I don't anticipate any real conflicts with #205, but I will be transitioning to `Management.ViewState` before I merge that PR.


apps/management/lib/management/view_state.ex
```diff
@@ -0,0 +1,62 @@
...
+        |> Brook.get!(unquote(collection), key)
+        |> Ok.ok()
+      catch
+        _, reason -> {:error, reason}
+      end
+
+      @impl true
+      def get_all do
+        unquote(instance)
+        |> Brook.get_all(unquote(collection))
```
jdenen, line 50: I'm still undecided if I like this returning the entire key/value view state or if I want to call `Brook.get_all_values/2` here.

NathanielScottStevens: I would do the latter. That's generally what I've found myself wanting when I call `get_all`.

apps/management/lib/management/view_state.ex
```diff
@@ -0,0 +1,62 @@
+defmodule Management.ViewState do
+  @moduledoc """
+  View state management behaviour and default implementation on a per-collection
+  basis. `ViewState` collections should track only one type of object.
+
+  ## Configuration
+
+  Configure `ViewState` with `use/2` macro.
```
jeffgrunewald, line 8: technically it's a `use/1` since it's looking for a keyword list.

apps/management/lib/management/view_state.ex
jeffgrunewald: get all values seems limiting but maybe we never find that use case where we need to unwrap the keys.

apps/management/lib/management/view_state.ex
```diff
@@ -0,0 +1,62 @@
...
+      @impl true
+      def persist(key, object) do
+        unquote(collection)
+        |> Brook.ViewState.merge(key, object)
+      end
+
+      @impl true
+      def get(key) do
+        unquote(instance)
+        |> Brook.get!(unquote(collection), key)
```
jeffgrunewald, line 41: why are you calling `Brook.get!` here and piping to `Ok.ok` but below you're calling `Brook.get_all` without the `!`?

apps/management/lib/management/view_state.ex
jdenen: First argument, `Management.ViewState`. Second argument, configuration.

apps/management/lib/management/view_state.ex
jdenen: So `Brook.get` and `Brook.get!` do the same thing. They either return `{:ok, _}` or raise an exception. Seems like a miss on Brook's part.

I use `get!` here to denote that an exception will be raised. `get` would hide that.

apps/service_gather/lib/gather/view_state/extractions.ex
```diff
@@ -0,0 +1,8 @@
+defmodule Gather.ViewState.Extractions do
```
jeffgrunewald, line 1: Do these need to be in separate files if they're essentially the same things and _only_ differ in the name of the collection?
This is reducing the boilerplate but it feels like a really low bar to reduce it a little further at least.

jdenen: Are you saying I should put the 3 modules in the same file? It's the same amount of code whether it's in one or three files. I don't really consider that reducing boilerplate.

apps/management/lib/management/view_state.ex
jdenen: I'm not going to change it in this PR, but I'm going to play with changing it in the next PR.

apps/service_gather/lib/gather/view_state/extractions.ex
jeffgrunewald: "boiler plate" may not be the best word to describe it but i think it's easier to lose track of how many of these modules there are if we need to make any changes, particularly if you add more to accommodate additional collections.

apps/management/lib/management/view_state.ex
jeffgrunewald: good point; wrong context.

apps/management/lib/management/view_state.ex
jeffgrunewald: `Brook.get!` _shouldn't_ be returning an :ok tuple. Under the hood `Brook.get!` is just calling `Brook.get` and unwrapping the tuple, but if you're using `Brook.get!` and getting back an {:ok, _} then there's a nested :ok tuple being stored.

jdenen: Yeah, sorry, it's the error path that's the same across `get` and `get!`. My thought process still applies though. I wanted to be explicit about that error behavior so I'm using `get!` for now.

apps/service_gather/lib/gather/event/handler.ex
```diff
@@ -6,37 +6,65 @@ defmodule Gather.Event.Handler do
...
+      ViewState.Sources.delete(key)
+    end)
+  end
+
+  def delete_destination(key) do
+    ViewState.Destinations.get(key)
+    |> Ok.map(fn dest ->
+      Destination.delete(dest)
+      ViewState.Destinations.delete(key)
+    end)
```
jeffgrunewald, line 68: i could see this being a common pattern; maybe it's worth moving these functions into the management library with the rest of the default implementations?

apps/service_gather/lib/gather/view_state/extractions.ex
jdenen: I'm not going to make this change in this PR, but I'm going to play around with it in the next few. I'm not convinced, but I get your point and want to see it in code.

apps/service_gather/lib/gather/event/handler.ex
jdenen: Very possible. We'll see how it looks as I convert more services.

PR #226, opened 2020-04-22
Broadcast and Persist use Management.ViewState

jdenen:
More of the same (per #222 and #224).


apps/service_broadcast/test/broadcast/event/handler_test.exs
```diff
@@ -68,33 +81,35 @@ defmodule Broadcast.Event.HandlerTest do
...
 
-      [transform: transform, load: load, delete: delete]
+      [transform: transform, load: load, delete: delete, key: key]
     end
 
-    test "deletes the transformation", %{load: load} do
+    test "deletes the transformation", %{key: key} do
       assert_async do
-        assert {:ok, nil} == Broadcast.Transformations.get(load.dataset_id, load.subset_id)
+        assert {:ok, nil} = ViewState.Transformations.get(key)
```
jeffgrunewald, line 89: Did you mean to convert this to a match assert instead of an equality assert?

jdenen: Yeah, I prefer the pattern match when working with ok/error tuples. I don't know why.

PR #227, opened 2020-04-24
Migrate service_acquire to use Management.ViewState

jdenen:
Also renames test directories that were inconsistent with their `lib/` counterparts.


apps/service_acquire/test/acquire/event/handler_test.exs
```diff
@@ -0,0 +1,76 @@
...
+  import Events, only: [transform_define: 0, load_start: 0, dataset_delete: 0]
+  import Definition, only: [identifier: 1]
+  alias Acquire.ViewState
+
+  @instance Acquire.Application.instance()
+
+  setup do
+    on_exit(fn ->
+      Brook.Test.clear_view_state(@instance, ViewState.Fields.collection())
+      Brook.Test.clear_view_state(@instance, ViewState.Fields.collection())
```
jeffgrunewald, line 12: is one of these intended to be `ViewState.Destinations.collection()`?

apps/service_acquire/lib/acquire/query/where/temporal.ex
```diff
@@ -28,6 +29,12 @@ defmodule Acquire.Query.Where.Temporal do
     |> and_clause()
   end
 
+  defp dictionary(dataset_id, subset_id) do
+    with {:ok, nil} <- identifier(dataset_id, subset_id) |> Acquire.ViewState.Fields.get() do
+      Ok.error("dictionary not found for #{dataset_id} #{subset_id}")
+    end
+  end
```
jeffgrunewald, line 36: is there a central way to share this boiler plate function with the `Where.Bbox` module?

apps/service_acquire/test/acquire/event/handler_test.exs
jdenen: Yes, nice catch.

PR #228, opened 2020-04-27
Decode Json arrays

ManApart:


apps/decoder_json/lib/decoder/json_arrays.ex
```diff
@@ -0,0 +1,32 @@
+defmodule Decoder.JsonArrays do
+  @moduledoc """
+  Defines a `Decoder` impl for arrays of JSON data.
```
jdenen, line 3: I think you need more here. How/why is this different than JSON lines?

ManApart: Done, how's that?

jdenen: LGTM

PR #238, opened 2020-05-01
Remove service_define

jdenen:
- We're not using it
- It's less than half-finished
- No one is going to be working on it... indefinitely
- It makes feedback times longer
- It makes `docker build` times longer
- We can always revert this commit if we want it again later


jdenen:
> could just be disabled in the build/deploy process
It's going to sit untouched. If I add a new pattern to all the services (like I just did) am I going to ignore Define (like I just did)? Am I going to do that indefinitely? I would rather delete it and preserve it in Git history than have unmaintained code.

> sets a pattern for how we would _like_ this work to be done
Not sure I entirely agree. While the typespec stuff is something we'd likely reuse, this code sets us down the ReactJS path that I'm not sure we ultimately want.

> Why not disable it from build and deploy and mark it as a "help wanted" to preserve that work
Because we're not going to maintain it. We're not going to update Define with the ViewState.Management pattern or any other new patterns moving forward. And if we're not going to do that, there's no reason to keep the code around.

> re-learn all the stuff that went into this and re-implement it from scratch
No one is going to re-implement Define without talking to you, me, or @bbalser first. I'm pretty sure we can point them in the right direction.

PR #242, opened 2020-05-04
Ability to extend hindsight with custom protocol implementations

jdenen:
If I want to add a custom, very specific-to-me `Source` implementation to Hindsight, I can bake it into an extending Docker image and Hindsight will pull it into the system.

```elixir
# my_plugins/foo.ex
defmodule Foo do
  defstruct [:bar, :baz]

  defimpl Source do
    # ...
  end
end
```
```Dockerfile
FROM inhindsight/hindsight:latest
COPY my_plugins/ plugins/
```
```
$ docker build -t myhindsight:extended .
```
All Hindsight services started in the `myhindsight/hindsight` image will compile and load plugins on startup. And so, the `Foo` destination is available for use throughout the system. This should work for any/all protocol implementations.

If a plugin file can't be compiled, applications will blow up immediately so users can't continue on the assumption that everything is OK. If no plugins exist or the directory is missing (local dev), Hindsight considers everything fine.

Some things to consider:
1. I'm not testing this through the Docker layer beyond manually.
2. Developing a custom impl will be a PITA since that code base will have no reference to our protocols without recreating them.
3. Deploying via Helm will require `image.repository` and `image.tag` to be set to the extended image values.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,3 +19,4 @@ ENV PORT=80
 EXPOSE ${PORT}
 WORKDIR /opt/app
 COPY --from=build /opt/app/_build/prod/rel/ .
+RUN mkdir plugins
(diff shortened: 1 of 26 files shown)


jdenen:
Still to do:
- [ ] README
- [ ] See if I can fix the first consideration (testing)

PR #243, opened 2020-05-07
Use Postgres for ViewState storage

jdenen:
Squashed and rebased #205. This will make a few more changes for consistency and minor fixes.


jdenen:
Keeping it a draft PR until the latest `brook_storage_postgres` changes are released.

apps/initializer/lib/initializer.ex
```diff
@@ -30,29 +31,47 @@ defmodule Initializer do
...
-        else
-          _ -> {:stop, "Supervisor not available", state}
+
+          _ ->
+            {:stop, "Supervisor not available", state}
         end
       end
 
+      @retry with: constant_backoff(100) |> take(10)
+      defp do_wait_for_supervisor(supervisor) do
```
jeffgrunewald: <nitpick> i don't think `do_wait_for_supervisor/1` has a corresponding `wait_for_supervisor/1` so maybe rename this to avoid confusion about additional callbacks that don't need to be implemented? </nitpick>

jeffgrunewald (approved):

apps/initializer/lib/initializer.ex
```diff
@@ -30,29 +31,47 @@ defmodule Initializer do
...
+          _ ->
+            {:stop, "Supervisor not available", state}
         end
       end
 
+      @retry with: constant_backoff(100) |> take(10)
+      defp wait_for_supervisor(supervisor) do
+        sup_pid = Process.whereis(unquote(supervisor))
+
+        case sup_pid do
```
jessie-morris, line 64: Sup Pid!

jessie-morris (approved):

jeffgrunewald (approved):

apps/hook_create_db/test/test_helper.exs
```diff
@@ -0,0 +1 @@
+ExUnit.start(capture_log: true)
```
jeffgrunewald, line 1: bless you.

apps/initializer/lib/initializer.ex
```diff
@@ -11,7 +11,8 @@ defmodule Initializer do
 
     quote location: :keep do
       use GenServer
-      use Retry
+      use Annotated.Retry
```
jeffgrunewald, line 14: worth it to upgrade the old dep and use the native functionality now?

jdenen: At some point, absolutely. With this PR, absolutely not.

apps/secret_store/lib/secret_store/environment.ex
```diff
@@ -0,0 +1,25 @@
...
+    |> System.get_env(default)
+  end
+
+  defp assemble_name(name, key) do
+    [name, key, SecretStore.environment()]
+    |> Enum.reject(&is_nil/1)
+    |> Enum.reduce("", fn
+      "", acc -> acc
+      val, "" -> val
+      val, acc -> "#{acc}_#{val}"
```
jeffgrunewald: This function seems repeated; could probably move it up into the behaviour or similar shared module.

apps/secret_store/lib/secret_store/aws.ex
```diff
@@ -0,0 +1,45 @@
...
+    |> Enum.reduce("", fn
+      "", acc -> acc
+      val, "" -> val
+      val, acc -> "#{acc}-#{val}"
+    end)
+  end
+
+  defp decode(%{"SecretString" => data}, nil, default), do: data || default
+
+  defp decode(%{"SecretString" => data}, key, default) do
```
jeffgrunewald, line 41: I assume you tested this with live data from AWS (even just once manually) to validate that the response coming from `ExAws.SecretsManager |> ExAws` is itself decoded to a map but that the value of the `data` variable is still just a json string? The test files only seem to use Placebo so not sure what we should be expecting from ExAws.

apps/secret_store/test/test_helper.exs
```diff
@@ -0,0 +1 @@
+ExUnit.start()
```
jeffgrunewald, line 1:
```suggestion
ExUnit.start(capture_log: true)
```

jeffgrunewald (approved):

PR #245, opened 2020-05-12
Remove helm chart from hindsight, this now lives at inhindsight/charts

jmorris-pillar:


scripts/install
```diff
@@ -57,11 +52,13 @@ function hindsight {
         local -r set_values="--set $vals"
     fi
 
-    deps
-
     info "Installing Hindsight"
-    helm upgrade ${release} ./helm \
+    helm repo add inhindsight https://inhindsight.github.io/charts
```
rucker, line 56: Is this the correct URL? It's 404 for me although I can reach https://github.com/inhindsight/charts

jmorris-pillar: it is, theres. is. an implicit index.yaml for chart repos so the url is https://inhindsight.github.io/charts/index.yaml

ManApart: Lol. I asked the exact same question on the last PR

PR #246, opened 2020-05-13
Backpressure worldwide

jeffgrunewald:
Removes the streaming resource consumption of a data source from Service Gather and adds a `shutdown/1` function to the `Source.Handler` behaviour to allow finite data sources to signal their end-of-feed status and allow the source to perform cleanup operations.


apps/service_gather/lib/gather/extraction/source_stream.ex
```diff
@@ -8,41 +8,103 @@ defmodule Gather.Extraction.SourceStream do
...
+        messages =
+          Enum.map(chunk, &lowercase_fields/1)
+          |> normalize(extract)
+
+        :ok = write_to_destination(extract.destination, context.assigns.destination_pid, messages)
+      end)
+    catch
+      _, e ->
+        warn_extract_failure(extract, e)
+        send(context.assigns.pid, {:extract_failed, e})
```
jmorris-pillar, line 33: Do we want to stop processing for all problems?  This seems like a situation where the dlq would be employed?

jmorris-pillar: I think i understand that u'll only realy ever land in catch when the destination call fails since everythign else just works without exceptions, but I'm curious if that is true

apps/service_gather/test/gather/extraction_test.exs
```diff
@@ -15,7 +15,7 @@ defmodule Gather.ExtractionTest do
   Temp.Env.modify([
     %{
       app: :service_gather,
-      key: Gather.Extraction,
+      key: Gather.Extraction.SourceStream.SourceHandler,
```
bbalser, line 18: Feels like SourceStream has no meaning in this module name now.

apps/service_gather/lib/gather/extraction/source_stream.ex
```diff
@@ -8,41 +8,104 @@ defmodule Gather.Extraction.SourceStream do
...
 
     getter(:dlq, default: Dlq)
+    getter(:app_name, required: true)
 
-    def handle_batch(batch, context) do
-      send(context.assigns.pid, {:source_batch, batch})
-      :ok
+    @impl Source.Handler
+    def handle_batch(batch, %{assigns: %{extract: extract}} = context) do
+      Decoder.decode(extract.decoder, [batch])
```
bbalser, line 22: Decoder was written to handle a stream of chunks, doesn't seem like this is true any more, should we refactor decoder to just handle a batch as well?

apps/service_gather/lib/gather/extraction/source_stream.ex
bbalser: I think we need to understand what are the scenarios in which an error is raised.

apps/service_gather/lib/gather/extraction/source_stream.ex
jeffgrunewald: yes? do we see a situation in which we would want decoder to be handling a stream at all given the problem that brought us here?
also, who is _we_ in this scenario refactoring Decoder?

apps/service_gather/test/gather/extraction_test.exs
jeffgrunewald: Agreed

apps/service_gather/lib/gather/extraction/source_stream.ex
```diff
@@ -8,41 +8,104 @@ defmodule Gather.Extraction.SourceStream do
...
+      end)
+
+      reason
+    end
   end
 
-  defp start_source(extract) do
-    {:ok, source_pid} = Source.start_link(extract.source, source_context(extract))
-    {extract.source, source_pid}
+  def start_source(extract, destination_pid) do
```
bbalser, line 101: this function like it fits better in Gather.Extraction than in the SourceHandler

apps/service_gather/lib/gather/extraction/source_stream.ex
jeffgrunewald: Effectively we were doing this already; if the extract were churning through a stream and failed to write it would kick the entire thing back to the `Extraction.extract` function which would retry a certain number of times and then kill the whole stream. At least now this is a more granular operation on the handling of each individual batch. We should only reach that scenario if for some reason our destination is completely unreachable. (also checked with @jdenen and he agreed the "shut it all down" approach was okay/best)

apps/service_gather/lib/gather/extraction/source_stream.ex
bbalser: I don't think so, I think with the new source protocol, it is unlikely we would go back to a stream implementation

apps/service_gather/lib/gather/extraction/source_stream.ex
jeffgrunewald: done

apps/service_gather/test/gather/extraction_test.exs
jeffgrunewald: Moved the only `SourceStream` function back up to `Extraction` and then got rid of the empty `SourceStream` wrapper module so now it's just `Gather.Extraction.SourceHandler`.

apps/protocol_source/lib/source/handler.ex
```diff
@@ -16,12 +16,14 @@ defmodule Source.Handler do
   @callback handle_message(map, Source.Context.t()) :: {:ok, map} | {:error, term}
   @callback handle_batch(list(map), Source.Context.t()) :: :ok
   @callback send_to_dlq(list(DeadLetter.t()), Source.Context.t()) :: :ok
+  @callback shutdown(Source.Context.t()) :: :ok
```
ManApart, line 19: FWIW I like this name. Makes sense.

apps/service_gather/lib/gather/extraction/source_stream.ex
jeffgrunewald: Implementing in a different card.

apps/service_gather/lib/gather/extraction.ex
```diff
@@ -28,28 +28,56 @@ defmodule Gather.Extraction do
...
   def handle_continue(:extract, %{extract: extract} = state) do
-    Logger.debug(fn -> "#{__MODULE__}: Started extraction: #{inspect(extract)}" end)
-
     case extract(extract) do
-      {:ok, _destination_pid} ->
-        Logger.debug(fn -> "#{__MODULE__}: Extraction Completed: #{inspect(extract)}" end)
-        Brook.Event.send(Gather.Application.instance(), extract_end(), "gather", extract)
-        {:stop, :normal, state}
+      {:ok, pid} ->
+        Logger.debug(fn -> "#{__MODULE__}: Started extraction: #{inspect(extract)}" end)
```
ManApart, line 33: So this was when it was complete, but now is when the extraction starts?

jeffgrunewald: Before the entire meat of an extract happened in that one `extract/1` function. When that function returned, the extract was done (for good or bad). Now `extract/1` just _starts_ the extract process and when that process finishes, it’s up to the shutdown callback to notify the extractor through the new handle_info callbacks.

apps/service_gather/lib/gather/extraction.ex
```diff
@@ -28,28 +28,56 @@ defmodule Gather.Extraction do
...
   defp extract(extract) do
     with {:ok, pid} <- start_destination(extract),
-         :ok <- do_extract(pid, extract) do
+         :ok <- start_source(extract, pid) do
       {:ok, pid}
     end
   end
 
+  def start_source(extract, destination_pid) do
+    {:ok, _source_pid} =
```
jdenen, line 75: I think we need to keep a handle on this pid. While it absolutely will not be useful for HTTP-related sources, we'll probably want it for Kafka sources. If we want to delete the dataset or shut it down later, we'll want to stop those source processes.

jdenen: I'm not suggesting you need to worry about the delete/shutdown scenario in this PR, just that we should track it in state.

apps/service_gather/lib/gather/extraction/source_stream.ex
```diff
@@ -1,58 +1,98 @@
...
+
+  @max_tries get_config_value(:max_tries, default: 10)
+  @initial_delay get_config_value(:initial_delay, default: 500)
+
+  getter(:dlq, default: Dlq)
+  getter(:app_name, required: true)
+
+  @impl Source.Handler
+  def handle_batch(batch, %{assigns: %{extract: extract}} = context) do
+    Decoder.decode(extract.decoder, [batch])
```
jdenen, line 21: Why wrap the batch in another list? Shouldn't a batch be a list? Should this be `List.wrap` instead?

jeffgrunewald: Good callout; in practice these batches were single messages so we were getting "protocol not implemented" warnings for the `Enumerable` but `List.wrap` is a better approach.

bbalser: Won't this go away once we refactor the Decoder protocol

jeffgrunewald: I don't want this to wait on the decoder so i like the wrap approach and removing it if necessary can be part of the final implementation of the decoder refactor.

apps/service_gather/lib/gather/extraction.ex
jeffgrunewald: Addressed; let us know what you think.

PR #248, opened 2020-05-14
DE-160: Remove requirement of ID on event structs

johntodd32:


ManApart:
We'll need to remove the ID from all events, so I think a good next step is to apply similar changes to the other events.

apps/definition_accept/lib/accept.ex
```diff
@@ -3,15 +3,14 @@ defmodule Accept do
   Accept defines the structure of an accept operation
   as performed by the receive service. Accept provides
   a reference to the dataset identifier for the data to
-  be received, an unique identifier for the accept operation
-  and human-readable name, as well as a destination
-  to write the data to.
+  be received, a human-readable name, and a destination
+  to write the data to. Upon validation a UUID is
```
ManApart, line 7: Good catch updating the documentation!

johntodd32: The description of the destination still needs work, but thanks.

apps/definition_accept/test/accept_test.exs
```diff
@@ -0,0 +1,41 @@
+defmodule AcceptTest do
```
ManApart, line 1: Nice work TDDing this and making nice, legible tests!

apps/definition_accept/test/support/fakes.ex
```diff
@@ -0,0 +1,38 @@
...
+    defimpl Accept.Connection do
+        def connect(_accept, _opts) do
+        end
+    end
+end
+
+defmodule FakeDestination do
+    defstruct start_link: nil, write: nil, stop: nil, delete: nil
+
+    def new!(_overrides \\ []) do
```
ManApart, line 17: we could probably remove this since we're not using it.
```suggestion
    def new!() do
```

apps/definition_accept/test/accept_test.exs
```diff
@@ -0,0 +1,41 @@
...
+            assert observed.id == "123-456"
+        end
+
+        test "overwrites existing id with generated one" do
+            allow UUID.uuid4(), return: "123-456"
+
+            input_with_id = Map.merge(input(), %{id: "an id"})
+            {:ok, observed} = Accept.new(input_with_id)
+            assert observed.id == "123-456"
+        end
```
ManApart, line 39: It might be nice to have a test that creates the expected struct and then does a pinned assert to make sure we are getting the struct we expect with the fields we expect.

apps/definition_accept/lib/accept.ex
```diff
@@ -47,6 +46,12 @@ defmodule Accept do
             subset_id: nil,
             destination: nil,
             connection: nil
+
+  def on_new(ac) do
```
ManApart, line 50: Seems like a nice hook for something like this. Once we've got it working for all the events, I think we should talk about possibly changing it so the macro `new` does this. That would be dryer, but I think this feels cleaner / less surprising, so I may still lean this way even after we've looked at the alternative.

apps/definition_accept/test/accept_test.exs
johntodd32: Thanks!

apps/definition_accept/lib/accept.ex
bbalser: I believe it would be relatively easy to add this functionality to Definition so all structs/events would get it for free, it could even support structs that don't have an id field.  This can be a refactor later, but I would consider it instead of updating all the events with this custom logic.  Let me know if you want details about how we can do this.

apps/definition_accept/test/support/fakes.ex
johntodd32: Done

apps/definition_accept/test/accept_test.exs
johntodd32: I reorganized the existing tests to make use of the setup feature of ExUnit, and I think it does what you are requesting. Check it out.

jdenen:
Squash and merge out all the JIRA references. There's no JIRA here! 🙃

apps/definition_accept/lib/accept.ex
ManApart: Yeah, if it's not too tricky to do and also would support structs that don't have an id field, I think this is exactly what we should do. I wanted to go the current route first so at the least we'd have _something_ in the case that we do another team change / priority pivot etc. But I definitely think we should go the route you mention if we can.

apps/definition_accept/test/accept_test.exs
```diff
@@ -0,0 +1,40 @@
...
+      input_with_id = Map.merge(input, %{id: "an id"})
+      assert {:ok, _} = Accept.new(input_with_id)
+    end
+
+    test "accepts an input that does not contain an id", %{input: input} do
+      assert {:ok, _} = Accept.new(input)
+    end
+
+    test "generates an id if not present", %{input: input, output: output} do
+      assert {:ok, output} = Accept.new(input)
```
ManApart, line 32: I believe you need to pin output, or it will match anything. Try making the output incorrect in the setup and re-running the tests. The ones that assert on output should fail.
```suggestion
      assert {:ok, ^output} = Accept.new(input)
```

apps/definition_accept/test/accept_test.exs
```diff
@@ -0,0 +1,40 @@
...
+    end
+
+    test "overwrites existing id with generated one", %{input: input, output: output} do
+      input_with_id = Map.merge(input, %{id: "an id"})
+      assert {:ok, output} = Accept.new(input_with_id)
```
ManApart, line 37: Same with this guy.

apps/definition_accept/test/accept_test.exs
ManApart: Looks good! You may need to pin output though so he doesn't always match.

johntodd32:
Borked some squashes and merges and needed a different approach to ids, so I'm abandoning this one.

PR #249, opened 2020-05-14
Decoders process batches of messages, not streams

rucker:


apps/decoder_json/lib/decoder/json.ex
```diff
@@ -14,14 +14,12 @@ defmodule Decoder.Json do
   defimpl Decoder do
     def lines_or_bytes(_t), do: :line
 
-    def decode(t, stream) do
-      stream
-      |> Enum.to_list()
+    def decode(_t, messages) do
+      messages
       |> List.flatten()
```
bbalser, line 19: Do we still need to flatten the list here?

apps/decoder_json/lib/decoder/json_arrays.ex
```diff
@@ -10,12 +10,8 @@ defmodule Decoder.JsonArrays do
...
     def lines_or_bytes(_t), do: :line
 
-    def decode(t, stream) do
-      Stream.map(stream, &decode_chunk/1)
-    end
-
-    defp decode_chunk(chunk) do
-      Enum.flat_map(chunk, &Jason.decode!/1)
+    def decode(_t, messages) do
+      Enum.flat_map(messages, &Jason.decode!/1)
```
bbalser, line 14: Same here, should this be an Enum.map

apps/decoder_json/lib/decoder/json.ex
jeffgrunewald: Not sure; our main reason for looking for feedback was wondering what assumptions we can make about the format of the data coming into these decoders? They definitely won't be streams any longer, but can we assume they are all complete and well-formed packets/payloads or could they be split along arbitrary lines by some means?

rucker: Looks like not. We can get this changed

apps/decoder_gtfs/test/decoder/gtfs_test.exs
```diff
@@ -16,20 +16,19 @@ defmodule Decoder.GtfsTest do
   end
 
   describe "Decoder" do
-    test "decodes gtfs into stream" do
-      input =
-        File.stream!("VehiclePositions.pb", [], 2048)
-        |> Stream.chunk_every(10)
+    test "decodes gtfs into batch of messages" do
+      decoder = Decoder.Gtfs.new!([])
```
jeffgrunewald, line 20: good catch on the empty list here

apps/decoder_gtfs/test/decoder/gtfs_test.exs
```diff
@@ -16,20 +16,19 @@ defmodule Decoder.GtfsTest do
...
+      input = File.read!("VehiclePositions.pb")
 
-      output = Decoder.decode(Decoder.Gtfs.new!(chunk_size: 2), input)
+      output = Decoder.decode(decoder, input)
 
       expected =
         File.read!("VehiclePositions.pb")
```
jeffgrunewald, line 26: just make this `input` as well and avoid double-reading the file.

apps/decoder_json/lib/decoder/json.ex
jeffgrunewald: probably safe to go ahead and remove this `List.flatten`

apps/decoder_json/lib/decoder/json_lines.ex
```diff
@@ -10,13 +10,9 @@ defmodule Decoder.JsonLines do
...
-    def decode(_t, stream) do
-      stream
-      |> Stream.map(&decode_chunk/1)
-    end
-
-    defp decode_chunk(chunk) do
-      Enum.map(chunk, &Jason.decode!/1)
+    def decode(_t, messages) do
+      messages
+      |> Enum.map(&Jason.decode!/1)
```
jeffgrunewald, line 15: for a single function call, i wouldn't bother with the pipe here (see above in json_arrays)

apps/decoder_csv/lib/decoder/csv.ex
```diff
@@ -25,25 +25,23 @@ defmodule Decoder.Csv do
...
-        {[parsed_chunk], %{acc | skip: false}}
-      end)
+    def decode(t, messages) do
+      if skip_line?(t) do
+        messages
+        |> Enum.drop(1)
+        |> Enum.map(fn message -> parse(message, t.headers) end)
+      else
+        Enum.map(messages, fn message -> parse(message, t.headers) end)
+      end
```
jeffgrunewald, line 35:
```suggestion
        messages
        |> handle_headers(t)
        |> Enum.map(fn message -> parse(message, t.headers) end)
    end
      
    def handle_headers(messages, %{skip_first_line: false}), do: messages
    def handle_headers(messages, _t) do
      unless Process.get(:have_skipped_headers) do
        Process.put(:have_skipped_headers, true)
        Enum.drop(messages, 1)
      else
        messages
      end
    end  
```

apps/decoder_csv/lib/decoder/csv.ex
```diff
@@ -26,24 +26,48 @@ defmodule Decoder.Csv do
...
 
     def decode(t, messages) do
-      if skip_line?(t) do
-        messages
-        |> Enum.drop(1)
-        |> Enum.map(fn message -> parse(message, t.headers) end)
-      else
-        Enum.map(messages, fn message -> parse(message, t.headers) end)
-      end
+      do_decode(t, messages, t.skip_first_line && !already_skipped?())
```
jeffgrunewald, line 29:
```suggestion
      do_decode(t, messages, t.skip_first_line and !already_skipped?())
```
This would have the same effect but would act as a short-circuit if `t.skip_first_line` evaluates to false and not bother with the second half of the clause.

apps/decoder_csv/lib/decoder/csv.ex
```diff
@@ -26,24 +26,48 @@ defmodule Decoder.Csv do
...
+
+    # def skip_line?(%{skip_first_line: false}), do: false
+    #
+    # def skip_line?(_t) do
+    #   unless Process.get(:have_skipped_headers) do
+    #     Process.put(:have_skipped_headers, true)
+    #     true
+    #   end
+    # end
+
```
jeffgrunewald, line 70: don't forget to delete these

apps/decoder_csv/lib/decoder/csv.ex
```diff
@@ -26,24 +26,48 @@ defmodule Decoder.Csv do
...
+      Enum.map(messages, &parse(&1, t.headers))
+    end
+
+    defp do_decode(t, messages, true = _should_skip?) do
+      mark_skipped()
+
+      messages
+      |> Enum.drop(1)
+      |> Enum.map(fn message -> parse(message, t.headers) end)
+    end
```
jeffgrunewald: Curious what @bbalser thinks about `do_decode` just outputting the messages unmodified if `_should_skip? = false` and `messagees |> Enum.drop(1)` if `_should_skip? = true` and then having the output of `do_decode` just piped to `Enum.map(fn message -> parse(message, t.headers end)` since lines 33 and 41 are identical.

rucker: I pushed a commit that does this. I dig it.

apps/decoder_gtfs/test/decoder/gtfs_test.exs
rucker: Sorry, I don't understand what you mean here

apps/decoder_csv/lib/decoder/csv.ex
bbalser: both && and "and" are short circuited.

apps/decoder_csv/lib/decoder/csv.ex
```diff
@@ -25,25 +25,25 @@ defmodule Decoder.Csv do
...
-      Enum.reverse(buffer)
+      Enum.drop(messages, 1)
+    end
+
+    defp already_skipped?() do
+      Process.get(__MODULE__, %{}) |> Map.get(:have_skipped_headers, false)
+    end
+
+    defp mark_skipped() do
+      Process.put(__MODULE__, %{:have_skipped_headers => true})
```
jmorris-pillar, line 46: fwiw, you can write the map here as

%{have_skipped_headers: true}

apps/decoder_csv/lib/decoder/csv.ex
```diff
@@ -25,25 +25,25 @@ defmodule Decoder.Csv do
...
     def lines_or_bytes(_t), do: :line
 
-    def decode(t, stream) do
-      stream
-      |> Stream.transform(%{skip: t.skip_first_line}, fn chunk, %{skip: skip} = acc ->
-        parsed_chunk = parse_chunk(chunk, t.headers, skip)
-        {[parsed_chunk], %{acc | skip: false}}
-      end)
+    def decode(t, messages) do
+      do_decode(messages, t.skip_first_line and !already_skipped?())
```
jmorris-pillar, line 29: I might pull t.skip_first_line and !already_skipped?() in a super small function which returned the same result and name it should_skip().  I'm not strongly opinionated about this as it seems like possibly too much, but i dont like a method being called with a `and` clause as part of one of the parameters

apps/decoder_csv/lib/decoder/csv.ex
```diff
@@ -25,25 +25,25 @@ defmodule Decoder.Csv do
...
+      |> Enum.map(&parse(&1, t.headers))
     end
 
-    defp parse_chunk(chunk, headers, skip) do
-      {buffer, _} =
-        Enum.reduce(chunk, {[], skip}, fn
-          message, {buffer, false} ->
-            {[parse(message, headers) | buffer], false}
+    defp do_decode(messages, false = _should_skip?), do: messages
```
bbalser, line 33: do_decode doesn't really make sense anymore since this function no longer decodes anything.

rucker: You're right. I'll see what changes I can make. Or maybe this last refactor took it too far?

bbalser: I think the last refactor works well, just need to update names to better represent purpose of the function.

jmorris-pillar: Yea i think do_decode is really "process_headers"

jmorris-pillar: or "ensure_the_headers_are_skipped_or_not_based_upon_what_the_user_asked_us_to_do_and_if_we_did_it_already"

apps/decoder_csv/lib/decoder/csv.ex
```diff
@@ -25,25 +25,29 @@ defmodule Decoder.Csv do
...
+    defp process_headers(messages, true = _should_skip?) do
+      mark_skipped()
 
-      Enum.reverse(buffer)
+      Enum.drop(messages, 1)
+    end
+
+    defp should_skip?(t) do
+      t.skip_first_line and !already_skipped?()
+    end
```
jeffgrunewald, line 43:
```suggestion
    defp should_skip?(skip_first_line) do
      skip_first_line and !already_skipped?()
    end
```

apps/decoder_csv/lib/decoder/csv.ex
```diff
@@ -25,25 +25,29 @@ defmodule Decoder.Csv do
...
     def lines_or_bytes(_t), do: :line
 
-    def decode(t, stream) do
-      stream
-      |> Stream.transform(%{skip: t.skip_first_line}, fn chunk, %{skip: skip} = acc ->
-        parsed_chunk = parse_chunk(chunk, t.headers, skip)
-        {[parsed_chunk], %{acc | skip: false}}
-      end)
+    def decode(t, messages) do
+      process_headers(messages, should_skip?(t))
```
jeffgrunewald, line 29:
```suggestion
      process_headers(messages, should_skip?(t.skip_first_line))
```

PR #250, opened 2020-05-18
DE-160: Remove requirement of ID on struct events

johntodd32:


apps/definition/lib/definition.ex
```diff
@@ -37,6 +37,7 @@ defmodule Definition do
       @impl Definition
       def new(%{} = input) do
         map = for {key, val} <- input, do: {:"#{key}", val}, into: %{}
+        map = Map.put_new(map, :id, UUID.uuid4())
```
jmorris-pillar, line 40: put_new_lazy might be more performant as it would only generate the UUID if the key isnt already present.

apps/definition/test/definition_test.exs
```diff
@@ -17,28 +18,47 @@ defmodule DefinitionTest do
...
+      fake_uuid = "fake_uuid"
+      allow(UUID.uuid4(), return: fake_uuid)
+      {:ok, [fake_uuid: fake_uuid]}
+    end
+
+    test "generates an id if not present on input", %{fake_uuid: fake_uuid} do
+      input = %{version: 2, bar: 9001}
+      assert {:ok, %Foo{version: 2, id: ^fake_uuid, bar: 9001, baz: nil}} = Foo.new(input)
+    end
+
```
jmorris-pillar, line 61: Do we need a test that it preserves id's when they are present?

apps/definition/lib/definition.ex
johntodd32: Done

apps/definition/test/definition_test.exs
johntodd32: Yes and done

apps/definition/lib/definition.ex
```diff
@@ -37,6 +37,7 @@ defmodule Definition do
       @impl Definition
       def new(%{} = input) do
         map = for {key, val} <- input, do: {:"#{key}", val}, into: %{}
+        map = Map.put_new_lazy(map, :id, fn -> UUID.uuid4() end)
```
jeffgrunewald, line 40:
```suggestion
        map =
          for {key, val} <- input, do: {:"#{key}", val}, into: %{}
          |> Map.put_new_lazy(:id, fn -> UUID.uuid4() end)
```

apps/definition/test/definition_test.exs
```diff
@@ -17,28 +18,52 @@ defmodule DefinitionTest do
     end
 
     def migrate(%__MODULE__{version: 1} = old) do
-      struct(__MODULE__, %{version: 2, bar: String.to_integer(old.bar)})
+      struct(__MODULE__, %{version: 2, id: old.id, bar: String.to_integer(old.bar)})
```
jeffgrunewald, line 21: this will cause an `KeyError` exception if you pass in a version of the struct that was created before an id was a required field and since it hasn't been "newed" into a struct yet there's not default value (like nil) to fall back on.

apps/definition/mix.exs
```diff
@@ -26,7 +26,11 @@ defmodule Definition.MixProject do
       {:ok, in_umbrella: true},
       {:norm, "0.10.4"},
       {:jason, "~> 1.1"},
+      {:uuid, "~>1.1"},
```
jeffgrunewald, line 29: would be interested to hear @bbalser 's take on this vs. `puid` as a more performant alternative.

apps/definition/lib/definition.ex
johntodd32: Done

apps/definition/test/definition_test.exs
johntodd32: I think I've addressed your concern about this, but let me know if this doesn't cut it.

apps/definition/mix.exs
jmorris-pillar: I benchmarked these and on initial go it did not seem faster.

iex(3)> Benchmark.measure(&UUID.uuid4/0)
0.039
iex(4)> Benchmark.measure(&UUID.uuid4/0)
0.031
iex(5)> Benchmark.measure(&UUID.uuid4/0)
0.027

vs

iex(3)> Benchmark.measure(&AlphanumPuid.generate/0)
0.044
iex(4)> Benchmark.measure(&AlphanumPuid.generate/0)
0.029
iex(5)> Benchmark.measure(&AlphanumPuid.generate/0)
0.038

jmorris-pillar: There might be ways to make a PUID spec that has the appropriate entropy, collision resistance, that is faster, but i would expect the gains to be marginal so I say we punt for now.

apps/definition/lib/definition.ex
```diff
@@ -36,7 +36,12 @@ defmodule Definition do
 
       @impl Definition
       def new(%{} = input) do
-        map = for {key, val} <- input, do: {:"#{key}", val}, into: %{}
+        map =
+          for {key, val} <- input,
+              do: {:"#{key}", val},
+              into:
+                %{}
+                |> Map.put_new_lazy(:id, fn -> UUID.uuid4() end)
```
jdenen, line 44: This is hard to read. The `for` loop executes completely before the result is piped to your `Map.put_new_lazy` call. It took me a minute to parse that. Elixir's `for` loops are only good as long as they can sit on one line (IMO). After that, you're better off using `Enum`.

```elixir
map =
  Enum.map(input, fn {key, val} -> {:"#{key}", val} end)
  |> Map.new()
  |> Map.put_new_lazy(:id, &UUID.uuid4/0)
```

johntodd32: FWIW the for loop used to be on one line, but the code formatter suggested the multiline format. I've changed it to your Enum.map function.

PR #253, opened 2020-05-22
DE-165: Transformation add timestamp

johntodd32:

### DIFF
--- a/apps/transformer/lib/transform/add_timestamp.ex
+++ b/apps/transformer/lib/transform/add_timestamp.ex
@@ -1,9 +1,19 @@
-defmodule Transform.AddTimestamp do
-  use Definition, schema: Transform.AddTimestamp.V1
+defmodule Transform.AddTimestampField do
+  @moduledoc """
+  `Transform.Step` impl to add a timestamp field to a dataset. Timestamps will be stored in ISO-8601 format.
+
+  ## Configuration
+  * `name` - Required. Field name or path to store the timestamp under. Ex: `"timestamp"` or `["timestamp"]`
+  """
+  use Definition, schema: Transform.AddTimestampField.V1
+
+  @type t :: %__MODULE__{
+    name: String.t() | [String.t()]
+  }
 
   defstruct [:name]
 
-  defimpl Transform.Step do
+  defimpl Transform.Step, for: __MODULE__ do
     import Dictionary.Access, only: [to_access_path: 1]
 
     def transform_dictionary(%{name: name}, dictionary) do
@@ -16,16 +26,17 @@ defmodule Transform.AddTimestamp do
     end
 
     def create_function(%{name: name}, _dictionary) do
-      Ok.ok(fn record -> Map.put(record, name, DateTime.utc_now() |> DateTime.to_iso8601()) end)
+      Ok.ok(fn record -> Map.put(record, name, DateTime.utc_now() |> DateTime.to_iso8601()) end |> Ok.ok())
     end
   end
 end
 
-defmodule Transform.AddTimestamp.V1 do
+defmodule Transform.AddTimestampField.V1 do
+  @moduledoc false
   use Definition.Schema
 
   def s do
-    schema(%Transform.AddTimestamp{
+    schema(%Transform.AddTimestampField{
       name: access_path()
     })
   end
--- a/apps/transformer/lib/transform/add_timestamp.ex
+++ b/apps/transformer/lib/transform/add_timestamp.ex
@@ -8,25 +8,33 @@ defmodule Transform.AddTimestampField do
   use Definition, schema: Transform.AddTimestampField.V1
 
   @type t :: %__MODULE__{
-    name: String.t() | [String.t()]
-  }
+          name: String.t() | [String.t()]
+        }
 
-  defstruct [:name]
+  defstruct name: nil,
+            description: ""
 
   defimpl Transform.Step, for: __MODULE__ do
     import Dictionary.Access, only: [to_access_path: 1]
 
-    def transform_dictionary(%{name: name}, dictionary) do
+    def transform_dictionary(%{name: name, description: description}, dictionary) do
       name_path = to_access_path(name)
-      put_in(dictionary, name_path, Dictionary.Type.Timestamp.new!(%{
-        name: name,
-        format: "%FT%TZ"
-      }))
-      |>Ok.ok()
+
+      put_in(
+        dictionary,
+        name_path,
+        Dictionary.Type.Timestamp.new!(%{
+          name: name,
+          description: description
+        })
+      )
+      |> Ok.ok()
     end
 
     def create_function(%{name: name}, _dictionary) do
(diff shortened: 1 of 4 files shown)


apps/transformer/test/transform/add_timestamp_test.exs
```diff
@@ -0,0 +1,32 @@
+defmodule Transform.AddTimestampFieldTest do
+  use ExUnit.Case
+  use Placebo
+
```
jmorris-pillar, line 4: Should we write a test // handle the case where the transform tries to create a date field with the same name as a field which already exists?

apps/transformer/lib/transform/add_timestamp.ex
```diff
@@ -0,0 +1,52 @@
+defmodule Transform.AddTimestampField do
+  @moduledoc """
+  `Transform.Step` impl to add a timestamp field to a dataset. Timestamps will be stored in ISO-8601 format.
+
+  ## Configuration
+  * `name` - Required. Field name or path to store the timestamp under. Ex: `"timestamp"` or `["timestamp"]`
```
jmorris-pillar, line 6: what is the difference between naming it "timestamp" vs ["timestamp"]

apps/transformer/test/transform/add_timestamp_test.exs
johntodd32: Sure. What should it do in that case?

apps/transformer/lib/transform/add_timestamp.ex
johntodd32: As I understand it, Transforms can accept either a string for a field name or a list of atoms for an access path. This is covering both possibilities.

apps/transformer/lib/transform/add_timestamp.ex
```diff
@@ -0,0 +1,52 @@
...
+          name: name,
+          description: description
+        })
+      )
+      |> Ok.ok()
+    end
+
+    def create_function(%{name: name}, _dictionary) do
+      Ok.ok(fn record ->
+        Map.put(record, name, DateTime.utc_now() |> DateTime.to_iso8601()) |> Ok.ok()
```
jeffgrunewald, line 36: are both of these `Ok.ok/1` necessary? It feels redundant to me unless I'm missing something.

apps/transformer/test/transform/add_timestamp_test.exs
jmorris-pillar: I dont know to be honest.  @jdenen ?

apps/transformer/lib/transform/add_timestamp.ex
johntodd32: As I understand it, the create_function method returns a function as part of a tuple with the OK atom. That function, when acting on a record, returns another tuple with the OK atom and the new record made by calling the function. This is the same pattern followed by the delete and move field transforms.

apps/transformer/lib/transform/add_timestamp.ex
jeffgrunewald: if it's a list of atoms then it should be `"timestamp"` or `[:timestamp]`

apps/transformer/lib/transform/add_timestamp.ex
jdenen: `create_function` returns a function in an `{:ok, fun}` tuple. Running that function returns its own `{:ok, return_value}` tuple.

I agree that it's a bit confusing, but that's the `Transformer.transform_function` type.

apps/transformer/lib/transform/add_timestamp.ex
johntodd32: Well, I copied the documentation from the delete and move field transformations, where the arguments are either strings or lists of strings. Should I change the @moduledoc in all three?

apps/transformer/lib/transform/add_timestamp.ex
jdenen: Would it be easier to parse if we re-arranged?

```elixir
fn record -> 
  now = DateTime.utc_now() |> DateTime.to_iso8601()

  Map.put(record, name, now)
  |> Ok.ok()
end
|> Ok.ok()
```

apps/transformer/lib/transform/add_timestamp.ex
jdenen: The list allows you to nest. We should probably differentiate the example to make the point.

`"timestamp"` and `["key1", "key2", "key3"]`

apps/transformer/test/transform/add_timestamp_test.exs
jdenen: I don't think we should write that test. I don't believe we're handling that case for the `MoveField` transformation either, and I'd rather we do it for both in another PR at the same time.

apps/transformer/lib/transform/add_timestamp.ex
bbalser: I believe this should be a NaiveDateTime.  During normalization all timestamps are converted to UTC and turned into a NaiveDateTime.to_iso8601.

apps/transformer/test/transform/add_timestamp_test.exs
bbalser: I think we should not worry about that problem in the transformations themselves, that feels like something define should be catching.  When the user says I want a transformation, the dictionary can be checked for field collision then.  The more checks and conditions we keep out of the pipeline, the faster it will be.

apps/transformer/lib/transform/add_timestamp.ex
johntodd32: I changed the documentation as Johnson suggested.

apps/transformer/lib/transform/add_timestamp.ex
johntodd32: Did the rearrangement and switched to NaiveDateTime.

jeffgrunewald: > Would it be easier to parse if we re-arranged?
> 
> ```elixir
> fn record -> 
>   now = DateTime.utc_now() |> DateTime.to_iso8601()
> 
>   Map.put(record, name, now)
>   |> Ok.ok()
> end
> |> Ok.ok()
> ```

Yes, I think that helps slightly but the primary confusion is the "nested" `:ok` tuples; thanks for clarifying.

apps/transformer/lib/transform/add_timestamp.ex
```diff
@@ -0,0 +1,56 @@
+defmodule Transform.AddTimestampField do
+  @moduledoc """
+  `Transform.Step` impl to add a timestamp field to a dataset. Timestamps will be stored in ISO-8601 format.
+
+  ## Configuration
+  * `name` - Required. Field name or path to store the timestamp under. Ex: `"timestamp"` or `["key1", "key2", "key3"]`
```
jmorris-pillar, line 6: Still being picky, if ["key1", "key2", "key3"] is a hierarchcicicacacal relationship can we say like. ["parent_field", "child_field"].  Maybe theres context in the main module i'm missing but i still dont understand how to use the second notation the way it is written.

PR #256, opened 2020-06-05
Add Aggregator service

rucker:


apps/service_aggregate/lib/aggregate/event/handler.ex
```diff
@@ -0,0 +1,33 @@
...
+      }) do
+    Logger.debug(fn ->
+      "#{__MODULE__}: Received event #{extract_start()}: #{inspect(extract)}"
+    end)
+
+    Aggregate.Feed.Supervisor.start_child(extract)
+
+    identifier(extract)
+    |> Aggregate.ViewState.Extractions.persist(extract)
+  end
```
jeffgrunewald, line 23: @bbalser do we want to comment this one out or delete it entirely before merging this in? This seems like it would be the only thing that could cause issues (even just resource contention) as it would trigger aggregate and profile to do the same thing on _any_ new dataset ingestion.

jeffgrunewald: maybe not; since it's not actually doing the profile/aggregate, just storing an extract in state.

jeffgrunewald (approved):

apps/platform_runner/mix.exs
```diff
@@ -24,19 +24,21 @@ defmodule PlatformRunner.MixProject do
 
   defp deps do
     [
+      {:benchee, "~> 1.0", only: [:test, :integration]},
```
rucker, line 27: The actual diff here is that we added `service_profile` and `service_aggregate` and sorted the list.

apps/platform_runner/mix.exs
```diff
@@ -24,19 +24,21 @@ defmodule PlatformRunner.MixProject do
...
-      {:divo, "~> 1.1"},
-      {:websockex, "~> 0.4.0"},
-      {:bypass, "~> 1.0", only: [:test, :integration]},
-      {:benchee, "~> 1.0", only: [:test, :integration]},
-      {:credo, "~> 1.3", only: [:dev]},
-      {:testing, in_umbrella: true, only: [:test, :integration]}
+      {:service_persist, in_umbrella: true},
+      {:service_profile, in_umbrella: true},
+      {:service_receive, in_umbrella: true},
+      {:testing, in_umbrella: true, only: [:dev, :test, :integration]},
```
rucker, line 40: Not sure if we made the correct change here but we were attempting to satisfy this error when running `mix deps.get`:

```
Dependencies have diverged:
* testing (../testing)
  the :only option for dependency testing

  > In mix.exs:
    {:testing, [env: :dev, path: "../testing", in_umbrella: true, only: [:test, :integration]]}

  does not match the :only option calculated for

  > In /Users/rick.ucker/code/hindsight/apps/service_aggregate/mix.exs:
    {:testing, [env: :dev, path: "../testing", in_umbrella: true, only: [:dev, :test]]}

  Ensure you specify at least the same environments in :only in your dep
** (Mix) Can't continue due to errors on dependencies
```

apps/service_aggregate/lib/aggregate/event/handler.ex
bbalser: I think it is okay to leave

bbalser (approved):

jessie-morris (approved):

PR #258, opened 2020-06-09
Json serde

bbalser:
Still working on cleaning up tests, but here are the names I came up with.  Look them over and if you want to change them, lets try to do that now before we have to worry about repercussions of changing them.


apps/decoder_csv/lib/decoder/csv.ex
```diff
@@ -22,7 +23,7 @@ defmodule Decoder.Csv do
             headers: nil,
             skip_first_line: false
 
-  defimpl Decoder, for: __MODULE__ do
```
jeffgrunewald, line 25: this requirement was just deprecated and is now implied, correct? this is still happening under the hood?

bbalser: yes, when calling defimpl in a module the "for:" is not needed

jeffgrunewald (approved):

jessie-morris (approved):

jdenen (approved):

jeffgrunewald (approved):

